PR #11, opened 2026-02-26
fix: robust OpenAI-compatible SSE streaming

Pibomeister:
## Summary

Fixes SSE streaming for OpenAI-compatible providers that deviate from the standard protocol. Primarily affects MiniMax and DeepSeek, but all changes are **provider-agnostic** — every OpenAI-compatible provider benefits.

- **Decouple `MessageDelta`/`MessageStop` from `usage` presence** — providers that send `finish_reason` without `usage` in the same chunk (MiniMax) now correctly terminate the stream
- **Emit `ContentBlockStart` on non-empty first chunks** — providers that send real text from the first SSE chunk (not an empty prefix) now trigger proper content block initialization
- **Filter `<think>...</think>` reasoning blocks** — chain-of-thought content embedded in the text field (MiniMax, DeepSeek) is stripped from both streaming and non-streaming responses, with cross-chunk boundary handling
- **Offset tool call indices by +1** — prevents index collision between text blocks (index 0) and tool use blocks in `block_states`, fixing silent tool call loss
- **Fall back to tiktoken estimate when provider omits usage** — context counter (`ctx:`) now shows accurate estimates when streaming API ignores `include_usage`
- **Emit `IntermediateText` for final text in multi-turn loops** — the last text response after tool call iterations now persists to the TUI instead of being silently dropped

## Provider Impact

| Provider | Before | After |
|----------|--------|-------|
| MiniMax | Empty responses, no ctx counter, reasoning leaked | Fully functional |
| DeepSeek | `<think>` blocks shown as text | Reasoning filtered |
| OpenAI / OpenRouter | Working | Unaffected (backward compatible) |
| Ollama / LM Studio | Working | More robust stream termination |

## Files Changed

- `src/brain/provider/custom_openai_compatible.rs` — streaming + non-streaming fixes
- `src/brain/agent/service.rs` — token fallback + IntermediateText fix

## Test Plan

- [x] `cargo test --lib` — 496 tests passing
- [x] `cargo build --release` — compiles clean
- [ ] Manual test: MiniMax text-only response
- [ ] Manual test: MiniMax tool calls + follow-up text
- [ ] Manual test: OpenAI/OpenRouter still works (no regression)

### DIFF
--- a/src/brain/agent/service.rs
+++ b/src/brain/agent/service.rs
@@ -41,21 +41,37 @@ pub type ApprovalCallback = Arc<
 #[derive(Debug, Clone)]
 pub enum ProgressEvent {
     Thinking,
-    ToolStarted { tool_name: String, tool_input: Value },
-    ToolCompleted { tool_name: String, tool_input: Value, success: bool, summary: String },
+    ToolStarted {
+        tool_name: String,
+        tool_input: Value,
+    },
+    ToolCompleted {
+        tool_name: String,
+        tool_input: Value,
+        success: bool,
+        summary: String,
+    },
     /// Intermediate text the agent sends between tool call batches
-    IntermediateText { text: String },
+    IntermediateText {
+        text: String,
+    },
     /// Real-time streaming chunk from the LLM (word-by-word)
-    StreamingChunk { text: String },
+    StreamingChunk {
+        text: String,
+    },
     Compacting,
     /// Compaction finished — carry the summary so the TUI can display it
-    CompactionSummary { summary: String },
+    CompactionSummary {
+        summary: String,
+    },
     /// Build completed — TUI should offer restart
-    RestartReady { status: String },
+    RestartReady {
+        status: String,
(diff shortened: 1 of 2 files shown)


adolfousier:
Thank you very much for your contribution @Pibomeister I appreciate that. 

I have just pushed major bug fixes for OpenAI Compatibles that conflict with this pull due to missing the new reasoning pipeline but I will cherry-pick the 5 valuable changes manually into the codebase to merge.

I will tag you as co-author. Coming up next.

Thank you

adolfousier:
Thank you very much for your contribution @Pibomeister I appreciate that. 

Cherry-picked all 6 substantive fixes into main (b778de8) with co-author credit. Had to adapt to coexist with the reasoning persistence pipeline we shipped in v0.2.34. 

Your changes are live. Rebuilding binaries now.

The rustfmt reformatting was skipped to keep the diff focused.

PR #14, opened 2026-03-03
feat: telegram allowlist hot-reload via TelegramState

Pibomeister:
## Summary

- Move `allowed_users` into `TelegramState` behind a `Mutex` for runtime hot-reload
- Seed the allowlist from agent config on bot startup

## Details

Adds `allowed_users: Mutex<HashSet<i64>>` to `TelegramState` with three new methods:
- `update_allowed_users()` — replace the allowed set (called on config reload)
- `is_user_allowed()` — check if a user ID is permitted
- `allowed_user_count()` — get the current count

On startup, `TelegramAgent::start()` now seeds the shared state from its config. A follow-up can wire the handler to use `TelegramState.is_user_allowed()` instead of its own Arc copy.

## Test plan

- [x] `cargo build --release` succeeds
- [x] `cargo test` passes
- [ ] Manual: verify Telegram bot starts and responds to allowed users
- [ ] Manual: verify config reload updates the allowlist without restart

### DIFF
--- a/src/channels/telegram/agent.rs
+++ b/src/channels/telegram/agent.rs
@@ -92,6 +92,11 @@ impl TelegramAgent {
                 return;
             }
 
+            // Seed the allowlist into shared state so it can be hot-reloaded later
+            self.telegram_state
+                .update_allowed_users(self.allowed_users.iter().copied().collect())
+                .await;
+
             tracing::info!(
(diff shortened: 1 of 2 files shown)


adolfousier:
Hey @Pibomeister, v0.2.45 dropped with your contribution, and extended the allowlist hot-reload to Discord and Slack as well. Really solid foundation you put together, made it easy to extend. Thank you.

PR #61, opened 2026-04-07
fix(sanitize): guard against out-of-bounds panic in redact_secrets and redact_command

anschmieg:
## What

Three bugs were fixed in `src/utils/sanitize.rs`:

### Bug 1: Panic in `redact_command()` (line ~95)

**Root cause:** `redact_command()` used `lower = result.to_lowercase()` to find case-insensitive pattern positions, then applied those positions to slice the original `result`. Unicode characters that expand on `to_lowercase()` (e.g. Turkish `İ` → `i̇`, 2→3 bytes) cause `match_pos` to exceed `result.len()`, panicking on `result[match_pos..]`.

**Fix:** Replaced the `to_lowercase()` approach with a new `find_case_insensitive()` helper that searches the original string directly using byte-level ASCII case comparison. Since all `COMMAND_SENSITIVE_PATTERNS` are pure ASCII, this avoids any Unicode expansion and ensures positions are always valid in the original string.

### Bug 2: Same panic in `redact_secrets()` key-prefix scan (line ~330)

Same `to_lowercase()` offset-mismatch bug in the known-key-prefix scan. The bounds-check guard added in the previous commit only fixed the panic (not the latent wrong-redaction risk), and only for this section.

**Fix:** Replaced `to_lowercase()` with `find_case_insensitive()` in both the key-prefix scan and the Bearer/Authorization scan (section 4). Added bounds guard to section 4 as defence-in-depth.

### Bug 3: Wrong redaction from offset misalignment

Copilot correctly noted that even when offsets were in-bounds, they could be *wrong* — pointing to the wrong character range. This would cause partial or missed secret redaction.

**Fix:** By never creating a lowered copy of the haystack string, `find_case_insensitive()` guarantees positions are always valid *and* correct.

## Changes

- `src/utils/sanitize.rs`: +133/-30 lines
  - New `find_case_insensitive()` helper — ASCII case-insensitive search on the original string, no `to_lowercase()`
  - `redact_command()`: uses helper instead of `to_lowercase()` + `find()`
  - `redact_secrets()` sections 1 and 4: same replacement
  - Bearer scan (section 4): added bounds check as defence-in-depth

## Testing

7 new regression tests covering Unicode-expansion scenarios:
- `redact_command_unicode_expansion_no_panic` — `İİİ...authorization` input
- `redact_command_unicode_expansion_api_key` — `İİİ...api_key=` input
- `redact_command_unicode_expansion_bearer` — `İİİ...bearer` input
- `redact_secrets_unicode_expansion_no_panic` — `İİİ...sk-proj-` input
- `redact_secrets_unicode_expansion_bearer` — `İİİ...bearer` input
- `redact_command_unicode_normal_text` — normal Unicode text unchanged
- `redact_secrets_unicode_normal_text` — normal Unicode text unchanged

> Note: tests require `fullfp16` CPU feature; compile in this environment blocked by a transitive `gemm-f16` dep. Tests are structurally identical to the existing `verify_fix.rs` standalone test that confirmed the panic scenario.

### DIFF
--- a/src/utils/sanitize.rs
+++ b/src/utils/sanitize.rs
@@ -84,27 +84,31 @@ fn redact_command(cmd: &str) -> String {
         }
     }
 
-    // Redact inline header values and query params (case-insensitive)
-    let lower = result.to_lowercase();
+    // Redact inline header values and query params (case-insensitive).
+    //
+    // All COMMAND_SENSITIVE_PATTERNS are pure ASCII. We use find_case_insensitive
+    // (below) instead of to_lowercase() to avoid the byte-offset mismatch problem:
+    // to_lowercase() can expand Unicode chars (e.g. 'İ' → 'i̇', 2→3 bytes), causing
+    // positions from the lowered string to misalign with the original string, which
+    // in turn causes wrong redaction or panic on slicing.
     for pattern in COMMAND_SENSITIVE_PATTERNS {
         let mut search_start = 0;
-        while let Some(pos) = lower[search_start..].find(pattern) {
-            let match_pos = search_start + pos + pattern.len();
-            // `match_pos` can exceed `result.len()` when `to_lowercase()` expanded
-            // the string (e.g. Turkish 'İ' → 'i̇'). Guard before slicing.
-            if match_pos > result.len() {
+        while let Some(abs_pos) = find_case_insensitive(&result[search_start..], pattern) {
+            let true_pos = search_start + abs_pos;
+            let after = true_pos + pattern.len();
+            if after > result.len() {
                 break;
             }
             // Find end of the secret: whitespace, quote, or end of string
-            let secret_end = result[match_pos..]
+            let secret_end = result[after..]
                 .find(['"', '\'', ' ', '&', '\n'])
-                .map(|p| match_pos + p)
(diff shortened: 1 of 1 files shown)


adolfousier:
Hey Anschmieg, really appreciate you catching this and taking the time to fix it. That panic in redact was nasty and your patch handles it cleanly. 

Thanks for making OpenCrabs better. Its merged now, pushing next release.

adolfousier:
> Amazing! Thank you for the quick response, I truly appreciate your effort on this project. After I was using OpenClaw and Hermes, OpenCrabs is the first agent I click with. I am glad to see how actively you are working on it and how quick it is catching up in terms of features. While I am not a software engineer and have never worked in Rust before, if there is a way for me to contribute to this project, I will happily do so.

Wow. This is so good to hear man, thank you for sharing that.

Yes there is, providing feedback.

If you want contribute with the code, you can use opencrabs, Crab is my best coding partner, it will help you too, it knows the code very well and can help you contribute directly, even push PRs and issues to GH through gh cli.

Whatever you prefer and yes rest assured I will keep building OpenCrabs full time and tackle any issue asap.

Huge thank you

PR #146, opened 2026-06-01
fix(file_extract): collapse double extension from forwarded messages

leshchenko1979:
Collapse trailing double extensions (e.g. 'file.docx.docx' → 'file.docx').
Telegram preserves the original filename from forwarded messages; if the
source already had an extension matching its MIME type, Telegram appends the
same extension again, breaking parse_document on the resulting .docx.docx file.

Add unit tests for the helper.

### DIFF
--- a/src/utils/file_extract.rs
+++ b/src/utils/file_extract.rs
@@ -111,6 +111,33 @@ pub fn is_text_mime(mime: &str) -> bool {
         )
 }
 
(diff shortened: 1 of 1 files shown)


adolfousier:
Merged, thanks Alexey! Solid fix for the double extension issue with good test coverage. I found and fixed a small logic bug in the rebuild step where it was reappending the extension after stripping it, but the core approach is sound.

PR #153, opened 2026-06-02
fix: escape single quotes in dynamic shell tool params

leshchenko1979:
## Problem

Parameter values containing single quotes (`'`) produce garbled output.

## Fix

Handle single quotes in parameter values to prevent output corruption.


adolfousier:
Merged. Clean fix, good test coverage. The POSIX `'\''` idiom is the right call here. Thanks Alexey.

PR #156, opened 2026-06-03
fix: replace hardcoded ~/.opencrabs/ paths with profile-aware opencrabs_home()

leshchenko1979:
Fixes three hardcoded `~/.opencrabs/` paths that bypass named profile isolation:

### Changes

1. **`src/utils/file_extract.rs`** — `save_to_temp()` used `dirs::home_dir().join('.opencrabs')` instead of `opencrabs_home()`, so temp files bypassed profile isolation.

2. **`src/brain/provider/codex_oauth.rs`** — `token_path()` same pattern: OAuth tokens always went to `~/.opencrabs/auth/codex.json` regardless of active profile.

3. **`src/brain/prompt_builder.rs`** — `push_known_paths()` hardcoded `'~/.opencrabs/'` in the LLM prompt for ALL paths. Logs/config/keys are global (use `base_opencrabs_dir()`), brain files and plans are per-profile (use `opencrabs_home()`). LLM now sees correct paths for the active profile.

### Discussion needed (issue pending)

- `src/brain/tools/subagent/status.rs` — subagent status dir uses hardcoded global path (might need per-profile isolation)
- `src/brain/tools/dynamic/loader.rs` — tools.toml is loaded from global path, profile-specific tools.toml is ignored by runtime

### DIFF
--- a/.codegraph/.gitignore
+++ b/.codegraph/.gitignore
@@ -1,0 +1,16 @@
+# CodeGraph data files
+# These are local to each machine and should not be committed
+
+# Database
+*.db
+*.db-wal
+*.db-shm
+
+# Cache
+cache/
+
+# Logs
+*.log
+
+# Hook markers
+.dirty
--- a/src/brain/prompt_builder.rs
+++ b/src/brain/prompt_builder.rs
@@ -635,17 +635,21 @@ fn push_skills_section(prompt: &mut String) {
 /// question stays out; the goal is "next time you're told 'check
 /// the logs' you don't grep .git/".
 pub(crate) fn push_known_paths(prompt: &mut String) {
-    prompt.push_str(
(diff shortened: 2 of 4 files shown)


adolfousier:
Hey Alexey, good catch on the hardcoded paths. Quick note: PRs should be atomic, this includes an unrelated .codegraph/.gitignore addition.

No worries, I will merge and clean up on our end. For future PRs: keep each one focused on a single concern. Thanks for the contribution!

leshchenko1979:
> Hey Alexey, good catch on the hardcoded paths. Quick note: PRs should be atomic, this includes an unrelated .codegraph/.gitignore addition.

Sorry for that - this rule is in my guidelines, but it just keeps falling out of the context window and disregarded. Will work on that.

PR #161, opened 2026-06-03
fix(team): add missing mark_awaiting_input in team agent spawn loop

leshchenko1979:
Team-created agents never called `mark_awaiting_input()` after completing a round, so `wait_agent` could never observe round-boundary progress — it kept polling until timeout, always reporting "still running" even though the agent had produced output and was waiting for follow-up input.

This mirrors the pattern in `spawn.rs` (spawn_agent): after `manager.update_output()`, flip to `AwaitingInput` state. This lets `wait_agent` return the round output immediately instead of blocking on task-join semantics (the agent task never terminates at a round — only on input/cancel).

Also added the same `tracing::info!` logs that `spawn.rs` has for consistency (round-complete and cancelled-while-waiting).

Fixes: team-created agents never completing from `wait_agent` perspective.

### DIFF
--- a/src/brain/tools/subagent/resume.rs
+++ b/src/brain/tools/subagent/resume.rs
@@ -235,6 +235,14 @@ impl Tool for ResumeAgentTool {
                 match result {
                     Ok(response) => {
                         manager.update_output(&agent_id_clone, response.content.clone());
+                        // Flip to AwaitingInput so wait_agent can observe
+                        // round-boundary progress — same pattern as
+                        // spawn.rs / team/create.rs. Without this the
+                        // resumed agent's task never terminates at a
+                        // round (only on input/cancel), so the old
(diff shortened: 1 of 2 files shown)


adolfousier (approved):
Approved. Clean fix that mirrors `spawn.rs:292-299` exactly — once a team-created agent finishes a round, `wait_agent` now sees the state transition instead of polling until timeout.

Pushed one follow-up commit on top: `resume.rs` had the same missing `mark_awaiting_input` call (the resumed agent's task never terminates at a round either, only on input/cancel). Same three-line shape mirroring spawn.rs. All three callers (`spawn`, `team/create`, `resume`) now share the pattern so a future state-machine refactor catches them together rather than leaving resume as the lone outlier.

Build + clippy (`--lib --bins --tests -- -D warnings`) clean. Thanks for the fix.

PR #200, opened 2026-06-15
fix(service): probe unit file scope instead of relying on euid

leshchenko1979:
## Problem

`cmd_service` uses `running_as_root()` to decide whether to call `systemctl` (system scope) or `systemctl --user` for all service operations. This breaks when the binary runs as root but units were installed as user-level — `systemctl restart opencrabs.service` fails with "Unit opencrabs.service not found" because the unit lives in `~/.config/systemd/user/`.

## Fix

Add `unit_scope()` which probes the filesystem to determine where the unit was actually installed:
- `/etc/systemd/system/{name}.service` exists → system scope
- `~/.config/systemd/user/{name}.service` exists → user scope
- Neither exists → falls back to `running_as_root()` (covers fresh install)

`Install` keeps using `running_as_root()` since that decides where to write a *new* unit.

## Affected operations

Start, Stop, Restart, Status, Uninstall — all now use `unit_scope()` instead of `running_as_root()`.

### DIFF
--- a/src/cli/commands.rs
+++ b/src/cli/commands.rs
@@ -1736,6 +1736,36 @@ fn running_as_root() -> bool {
     unsafe { libc::geteuid() == 0 }
 }
 
+/// Core probe logic with explicit paths for testability.
+/// Returns `true` for system scope, `false` for user scope.
+#[cfg(target_os = "linux")]
+fn unit_scope_probe(
+    systemd_name: &str,
+    system_dir: &std::path::Path,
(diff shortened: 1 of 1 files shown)


adolfousier (approved):
Reviewed: probes the actual unit-file location (system vs user) and falls back to running_as_root() safely when neither exists. CI green. Merging; relocating the tests to src/tests/ and hardening user-home detection under sudo (SUDO_USER) in follow-up commits.

PR #208, opened 2026-06-19
fix: return brief confirmation on /compact instead of raw compaction summary

leshchenko1979:
## Problem

The manual `/compact` command returns the full internal compaction summary as the response. This summary is designed as LLM context, not for human consumption — it reads like raw prompt garbage to the user.

Auto-compaction works correctly because it injects the summary as context and lets the LLM generate a natural response. The manual path bypasses the LLM entirely and dumps the raw summary verbatim.

## Fix

Return a clean confirmation message with token count and usage percentage instead:

```
✅ Context compacted — now at 72% (58320 tokens).
```

The full summary is still persisted to DB for agent continuity (the LLM sees it on the next turn).

## Changes

- `src/brain/agent/service/tool_loop.rs`: Replace `content: summary` with a brief confirmation in the `is_manual_compact` branch. Updated comments to clarify the purpose.

### DIFF
--- a/src/brain/agent/service/tool_loop.rs
+++ b/src/brain/agent/service/tool_loop.rs
@@ -723,8 +723,8 @@ impl AgentService {
             .await
             .map_err(|e| AgentError::Database(e.to_string()))?;
 
-        // Manual /compact: force compaction and return summary directly — no second LLM call.
-        // The summary already contains next steps and follow-ups, so it IS the response.
+        // Manual /compact: force compaction, persist summary to DB, return a brief
+        // confirmation to the user. The full summary is for the agent, not the user.
         if is_manual_compact {
             match self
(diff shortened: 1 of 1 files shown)


adolfousier:
Merged, thanks Alexey. Also added a brief continuation prompt to the context after manual compact so the agent knows to resume from the summary. Matches auto-compaction behavior but keeps it short. Full POST-COMPACTION PROTOCOL stays for auto only.

adolfousier:
Merged, thanks Alexey. Solid fix as always.

What we did on top: added a Manual variant to CompactionKind with a short resume hint (one sentence, not the full POST-COMPACTION PROTOCOL). After manual /compact, the LLM now gets a brief continuation prompt in context matching how auto-compaction works. Also added a test to verify Manual is actually brief.

All CI green, pushed to main.

PR #511, opened 2026-07-11
feat: add extra_headers support for custom providers

adi805:
## Summary

Adds an optional `extra_headers` field to `ProviderConfig` that injects custom HTTP headers into every request to custom providers.

## Problem

Some API gateways (e.g. AgentRouter) validate client identity via User-Agent headers. OpenCrabs custom providers had no way to set custom headers — the runtime `OpenAIProvider` already supported `with_extra_headers()` but it wasn't exposed to TOML config.

## Solution

1. **`src/config/types.rs`**: Added `extra_headers: Option<BTreeMap<String, String>>` to `ProviderConfig`
2. **`src/brain/provider/factory.rs`**: Read `extra_headers` from config in `configure_openai_compatible()` and pass to `with_extra_headers()`

## Config Example

```toml
[providers.custom.agentrouter]
enabled = true
base_url = "https://agentrouter.org/v1"
default_model = "glm-5.2"
models = ["glm-5.2", "gpt-5.5"]
extra_headers = { User-Agent = "openclaw" }
```

## Testing

Verified with AgentRouter (requires `User-Agent: openclaw` header). Without the header, requests return 401 `unauthorized_client_error`. With it, both `glm-5.2` and `gpt-5.5` return 200 OK.

## Notes

- Field is optional — existing providers unaffected
- Uses `BTreeMap<String, String>` for TOML inline table syntax support
- Follows existing pattern (OpenRouter has hardcoded headers for its specific needs)

### DIFF
--- a/src/brain/provider/anthropic.rs
+++ b/src/brain/provider/anthropic.rs
@@ -44,6 +44,12 @@ pub struct AnthropicProvider {
     /// User override from `providers.anthropic.context_window` in config.toml.
     /// When set, becomes the compaction budget (overrides agent.context_limit).
     configured_context_window: Option<u32>,
+    /// Custom base URL override (e.g. for AgentRouter gateways).
+    /// When set, replaces `ANTHROPIC_API_URL` and `ANTHROPIC_MODELS_URL`.
+    base_url: Option<String>,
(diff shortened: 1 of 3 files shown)


adi805:
Superseded by combined PR

PR #762, opened 2026-07-25
fix(brain): widen TOOLS.md proactive load trigger

leshchenko1979:
## Problem

TOOLS.md is a contextual (on-demand) brain file loaded via `load_brain_file`. Its proactive load trigger in `prompt_builder.rs` was:

> "Working with environment-specific tool configs → load TOOLS.md"

This trigger is too narrow to match real task patterns. **Log analysis (Jul 21–25) shows 0 organic TOOLS.md loads across 198 total `load_brain_file` calls over 5 days.** The file is effectively dead weight.

TOOLS.md contains critical operational knowledge that prevents recurring tool failures:

| Content | Example |
|---------|---------|
| Tool failure patterns & fixes | `telegram_send` JSON errors, `hashline_edit` 100% failure rate, `follow_up_question` option length |
| Platform formatting rules | Telegram 4096 char limit, Discord embeds, WhatsApp markdown subset |
| Cron semantics | Cold-start sessions, self-bootstrap context, `thread_id` requirement |
| Agent tool ID hygiene | Full UUID required, verify before close/resume |
| Phantom tool warnings | `read`, `cat`, `find`, `rebuild` do not exist |
| Provider reliability | Dead providers (HTTP 410), retry policies |
| Dynamic tools format | `$OPENCRABS_PARAMS`, `tools.toml` schema |

None of this matches "environment-specific tool configs".

## Fix

Widen the trigger to cover the actual content:

```
- Using non-core tools (Telegram, cron, browser, agents, dynamic tools),
  debugging tool failures, sending channel messages, or scheduling work
  → load TOOLS.md
```

## Evidence

```
Date        load_brain_file total   TOOLS.md loads
Jul 21      17                      0
Jul 22      63                      0
Jul 23      11                      0
Jul 24      42                      0
Jul 25      65                      6 (all from investigation sessions)
```

## Notes

- No cargo available on this host — change is a string literal in `prompt_builder.rs`, no logic change.
- The AGENTS.md pointer section ("Tool quirks → TOOLS.md") already exists but is passive prose in an already-long always-loaded file. The proactive trigger is the primary discovery mechanism.

PR #763, opened 2026-07-25
feat(harness): proactive brain file hints in tool errors and tool_search

leshchenko1979:
## Problem

TOOLS.md is effectively dead weight — 0 organic loads across 198 `load_brain_file` calls, Jul 21–25. The model never loads it because:
1. The proactive trigger in the system prompt is too narrow ("environment-specific tool configs")
2. AGENTS.md pointers are passive prose
3. When a tool call fails, the error string carries zero guidance — no `tool_search` hint, no TOOLS.md pointer

The model's only option is to remember a rule from 20 turns ago. Weak models don't.

## Solution — harness-driven, not model-driven

The harness should do the work, not the model. When a tool call fails or a tool is discovered, the harness searches brain files (TOOLS.md, AGENTS.md, CODE.md, SOUL.md) via the **existing FTS5 index** and injects relevant snippets into the tool result. The model just reads them inline — no action required.

**Why FTS, not heading-based injection:**
- The `qmd` FTS5 engine is already loaded for memory search (sub-millisecond, BM25)
- Brain files are already indexed at startup (`memory::index::BRAIN_FILES`)
- Content-based — works with any TOOLS.md formatting, no `## tool_name` convention required
- Covers ALL brain files, not just TOOLS.md

## Changes

| File | Change |
|------|--------|
| `src/brain/hints.rs` (new, 156 lines) | `hints_for_tool(tool_name, error_context)` — FTS search + snippet formatting |
| `src/memory/search.rs` (+63) | `search_brain()` — FTS-only over brain collection, no vector/embedding overhead |
| `src/memory/mod.rs` | Make `COLLECTION_BRAIN` `pub(crate)`; re-export `search_brain` |
| `src/brain/mod.rs` | Register `hints` module |
| `src/brain/agent/service/tool_loop.rs` (+17) | Inject hints into "Tool not found" (line 5149) and "Tool execution error" (line 5557) |
| `src/brain/tools/tool_search.rs` (+7) | Inject hints into discovery results |

## Example

Before:
```
Tool not found: tg_send_message
```

After:
```
Tool not found: tg_send_message

─── relevant notes ───
• **TOOLS.md** — `telegram_send` is the canonical name; `tg_*` is an abbreviation that the
  self-heal layer maps. Use `telegram_send` directly to avoid the extra round-trip.
```

## Design notes

- **FTS-only, not hybrid**: Harness hints are sub-millisecond critical. Skip vector embeddings and RRF.
- **Over-fetch + filter**: Search returns `n*3`, filter to brain collection, trim to `n`. Defense in depth.
- **OnceCell guard**: If the FTS store is unavailable, don't retry on every tool call.
- **No model action required**: The model reads the hints as part of the tool result. No new rule to remember.
- **Max 2 snippets, 900 chars total**: Keeps tool results lean.

## Testing

- Unit tests in `hints.rs`: `truncate_at_word`, `format_hints`
- FTS integration is covered by existing `memory_store_test.rs`
- No new cargo deps — uses existing `qmd` crate

## Related

- Companion to #762 (widens the TOOL LIFECYCLE trigger in the system prompt)
- Addresses the "model never loads TOOLS.md" problem from the user's analysis

🤖 Generated with [OpenCode](https://opencode.ai)

Co-Authored-By: OpenCode <noreply@opencode.ai>

### DIFF
--- a/src/brain/agent/service/tool_loop.rs
+++ b/src/brain/agent/service/tool_loop.rs
@@ -2,6 +2,7 @@ use super::builder::AgentService;
 use super::types::*;
 use crate::brain::agent::context::AgentContext;
 use crate::brain::agent::error::{AgentError, Result};
+use crate::brain::hints::hints_for_tool;
 use crate::brain::provider::{ContentBlock, LLMRequest, LLMResponse, Message};
 use crate::brain::tools::ToolExecutionContext;
 use crate::services::{MessageService, SessionService};
@@ -5147,10 +5148,16 @@ impl AgentService {
                         } else {
                             // Tool not found, skip approval
                             let err = format!("Tool not found: {}", tool_name);
-                            tool_outputs.push((false, err.clone()));
+                            // Harness-driven brain file hints: search TOOLS.md
+                            // etc. for relevant context about the missing tool.
+                            let mut content = err.clone();
+                            if let Some(hint) = hints_for_tool(&tool_name, None).await {
+                                content.push_str(&hint);
+                            }
+                            tool_outputs.push((false, content.clone()));
                             tool_results.push(ContentBlock::ToolResult {
                                 tool_use_id: tool_id,
-                                content: err,
+                                content,
                                 is_error: Some(true),
                             });
                             continue;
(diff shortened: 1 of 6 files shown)


leshchenko1979:
Already implemented

PR #818, opened 2026-07-26
feat(desktop): OpenCrabs desktop GUI (Dioxus WASM + Tauri 2, beta-stable)

moneyacademyKE:
## What this adds

A new, **self-contained `desktop/` subproject**: a native macOS GUI for OpenCrabs — a **Dioxus 0.7 WASM frontend running in a Tauri 2 webview**, talking to the **in-process `opencrabs` crate** over a typed IPC bridge. No changes outside `desktop/`.

## Status

**Beta-stable.** The frontend mounts, IPC is verified end-to-end (sessions / tools / usage / diagnostics), and releases are gated by a reproducible ladder + a native macOS smoke.

## Layout

| Path | Contents |
|---|---|
| `desktop/src/` | Dioxus WASM frontend — `app.rs`, `models.rs`, `bridge.rs`, `public/css/app.css` |
| `desktop/src-tauri/` | Native Tauri 2 shell + typed IPC commands (`chat`, `session`, `cron`, `tools`, `files`, `brain`, `config`, `usage`, `diagnostics`, …) backed by `opencrabs` |
| `desktop/docs/adr/` | Architecture decisions (build via `dx`; native mount verification) |
| `desktop/scripts/` | `release-verify.sh` (fail-closed gate ladder) + `native-smoke.sh` (macOS launch/mount smoke) |
| `desktop/README.md`, `STABILITY-EVIDENCE.md`, release notes | Docs |

## Build

```sh
rustup toolchain install 1.91.0            # rust-toolchain.toml pins it
cargo install dioxus-cli --version 0.7.9   # the `dx` CLI builds the Dioxus frontend
cargo install tauri-cli --version "^2"
cd desktop && cargo tauri build             # frontend via `dx build --release`
```

## Verification

- `desktop/scripts/release-verify.sh` runs **fmt / clippy / test for both crates, `dx build --release`, the Tauri bundle, and a native macOS smoke** — all green.
- **Compiles against this branch's base**: `cargo check` of the desktop native crate passes cleanly against `opencrabs` **v0.3.75** (it uses stable APIs present in `main`).

## Compatibility note

Developed on a fork; rebased cleanly onto `main` here. Depends on `opencrabs` via a path dependency (`path = "../.."`) with no version pin, so it tracks your `main`. If you'd prefer a version-pinned dep or a workspace inclusion, happy to adjust.

## Known limitations (beta)

- **macOS Apple-Silicon only** (arm64 `.app`/`.dmg`).
- **Unsigned** — Gatekeeper will prompt on first launch (right-click → Open). Signing/notarization runbook is in `desktop/docs/release-runbook.md`.
- In-app **update-install** and **voice** are explicitly unsupported and labeled as such in the UI.
- Not a production-stable release yet — beta.

## Scope

`desktop/` only — deliberately no changes to the CLI/TUI or repo-root files, to keep review focused and avoid any conflict with `main`.


adolfousier (changes requested):
Reviewed the full diff (114 files), read the security-critical command modules, and compiled the branch against current main. Overall this is high-quality work: allowlisted config writes with zero API-key exposure, real workspace path containment, escaped rendering (the old XSS critical is gone), honest fail-closed stubs, and an evidence doc that actually cites build hashes. Notes below so we can land this properly.

## Required (blocks merge)

**1. Does not compile against current main.** The branch base is v0.3.75-era and main moved since. `cargo check` in `desktop/src-tauri` against current main fails with exactly one error:

```
error[E0063]: missing field `include_subagents` in initializer of `SessionListOptions`
  --> desktop/src-tauri/src/commands/session.rs:80
```

`include_subagents` was added in 0f44a49a (feat(sessions): hide and expire spawned sub-agent sessions). Fix:

```rust
SessionListOptions {
    include_archived: false,
    include_subagents: false,
    limit: 40,
    offset: 0,
    query: None,
}
```

Verified: with that one line applied, the full gauntlet (`cargo clippy --locked --lib --bins --tests --all-features -- -D warnings`) passes with zero warnings against current main, so nothing else is hiding behind it. The rebase will also refresh `desktop/src-tauri/Cargo.lock` (ui crate lock unaffected).

**2. CI has never run on this PR.** Run 30221910230 has been sitting in `action_required` since Jul 26 (first-time-contributor gate). It needs maintainer-side approval before checks will report at all. Heads up for after: even then, ci.yml does not build `desktop/` (it is not a root workspace member), so before merge we need a path-filtered CI job that builds the frontend first and then checks/tests `desktop/src-tauri` (a bare `cargo check` fails in `generate_context!` without the dist dir).

## Should fix (non-blocking, but please address)

1. **CSP vs docs mismatch.** AUDIT.md claims the CSP no longer allows `unsafe-inline`, but tauri.conf.json still carries `script-src 'self' 'unsafe-inline' 'unsafe-eval' 'wasm-unsafe-eval'` plus a broad `connect-src ... https:`. With `withGlobalTauri: true`, either drop `unsafe-inline` and scope `connect-src`, or correct the doc.
2. **approve_tool writes the global policy.** One tool gesture rewrites the repo-wide `agent.approval_policy` via apply_approval_policy. The current UI only reaches on-request/manual (auto-always is unreachable), but approving a single tool silently changing the global agent policy is a trap. Make it explicit in the UI or move to per-tool semantics.
3. **scrubbed_key panic.** `&trimmed[..4]` byte-slices and panics on non-ASCII keys. Use `chars()`.
4. **add_dynamic_tool clobbers tools.toml.** Rewriting via `toml::to_string_pretty` drops user comments and unmodelled fields. Merge into the existing document with toml_edit instead.
5. **Onboarding gap.** save_onboarding_config enables the provider but never sets `agent.default_provider`.
6. **File browser scope.** Workspace root is the process CWD and TEXT_EXTS includes `env`, so `.env` files are readable through the preview. Fine if deliberate for an owner-operated local tool, but confirm it is a decision and note it in the doc, especially while the CSP is still loose.
7. **Hygiene.** `desktop/src-tauri/gen/schemas` is generated output and should be gitignored. The `$schema` URL in tauri.conf.json points at a personal fork.

## Verified claims (for the record)

- XSS critical fixed: escaped interpolation at app.rs:1211, zero dangerous_inner_html in the frontend. Verified in code.
- File access containment: `..` rejection + canonicalize + starts_with, binary detection, 256KB cap. Verified.
- Config writes allowlisted via SAFE_CONFIG_KEYS; get_config exposes only has_api_key. Verified.
- Stubs (voice, update install) are honest and fail-closed. Verified.
- "Beta-stable" is backed by an evidence doc (build hash, sha256, dmg fresh-install, smoke suite) with limits openly stated (unsigned build, dx CLI dependency, no upgrade tests).

Strong contribution. Item 1 is one line, then we are close.

Issue #33, opened 2026-03-07
OpenCrabs cron jobs don't execute in daemon mode


opryshok:
@adolfousier please check.

Found the issue. Looking at the code:

Evidence:
• Cron scheduler spawned exists in binary
• src/cron/scheduler.rs exists
• Jobs are in DB with next_run_at timestamps
• But no cron execution logs appear

Root Cause: The cron scheduler appears to only work in TUI mode, not in daemon mode. The daemon spawns but doesn't run the cron schedule.

Steps to reproduce:
1. Start OpenCrabs in daemon mode: opencrabs daemon
2. Create cron job using cron_manage or add. Wait for scheduled time ( to DB
3e.g., 1 hour)
4. Cron job never executes

Expected behavior:
• Cron jobs should execute on schedule in daemon mode

Actual behavior:
• Jobs are saved in database with correct next_run_at timestamps
• No execution occurs
• No error logs in journalctl

Environment:
• OpenCrabs version: v0.2.59
• Mode: daemon
• Platform: Linux (systemd)

Database evidence:
SELECT * FROM cron_jobs;
-- Shows jobs with enabled=1, valid next_run_at timestamps
-- But last_run_at stays empty

PR #938, opened 2026-08-04
fix(db): make migration 33 idempotent for pre-existing provider column (#937)

### LINKED ISSUE #937: fix(db): migration 33 fails with 'duplicate column name: provider' if provider column already exists
## Problem Description

When running OpenCrabs on a database where `tool_executions` already contains the `provider`, `model`, or `duration_ms` columns (e.g. from an intermediate build, dev run, or interrupted migration attempt), migration `#33` (`20260731000001_add_analytics_events.sql`) fails during startup with:

```text
Error: Failed to run database migrations: rusqlite_migration error while executing query:
'ALTER TABLE tool_executions ADD COLUMN provider TEXT;
ALTER TABLE tool_executions ADD COLUMN model TEXT;
ALTER TABLE tool_executions ADD COLUMN duration_ms INTEGER;':
duplicate column name: provider: Error code 1: SQL logic error
```

Since the DDL statement fails, `rusqlite_migration` does not update `PRAGMA user_version` to `33`, causing every subsequent startup of `opencrabs` / `opencrabs daemon` to fail repeatedly.

## Expected Behavior

Migration `#33` should execute idempotently. If `tool_executions` already contains the `provider` column, the migration runner should safely advance the migration version to 33 without crashing the application.

## Proposed Fix

Add a pre-migration probe in `src/db/database.rs` for migration `33`. If `tool_executions` already contains the `provider` column when `user_version < 33`, safely execute the new table creations (`phantom_events`, `streaming_recoveries`, `brain_verify_events`) and advance `user_version` to `33`.

moneyacademyKE:
## Summary

Fixes #937.

When upgrading or launching OpenCrabs on a database where `tool_executions` already contains the `provider`, `model`, or `duration_ms` columns from a prior build or partial migration attempt, migration `#33` (`20260731000001_add_analytics_events.sql`) fails with:

```text
duplicate column name: provider: Error code 1: SQL logic error
```

## Solution

In `src/db/database.rs`, add a pre-migration probe before `migrations.to_latest(conn)`. If `user_version < 33` and `tool_executions` already contains `provider`, the database runner safely creates the missing Mission Control analytics tables (`phantom_events`, `streaming_recoveries`, `brain_verify_events`) and advances `user_version` to `33` to prevent non-idempotent `ALTER TABLE` crashes.

## Verification

- `cargo check --bin opencrabs` passed cleanly.
- Tested locally on SQLite database with pre-existing columns — successfully advanced migration version to 33 and launched OpenCrabs without errors.

### DIFF
--- a/src/db/database.rs
+++ b/src/db/database.rs
@@ -297,6 +297,41 @@ impl Database {
                     conn.pragma_update(None, "user_version", Self::MIGRATION_COUNT as i64)?;
                 }
 
+                // Defensive check for migration 33 (add_analytics_events):
+                // If tool_executions table already contains provider column from prior schema state,
+                // ensure migration does not fail with duplicate column name error.
+                if user_version < 33 {
+                    let has_provider: bool = conn
+                        .prepare("PRAGMA table_info(tool_executions)")?
+                        .query_map([], |r| r.get::<_, String>(1))?
+                        .filter_map(Result::ok)
+                        .any(|col| col == "provider");
+
+                    if has_provider {
+                        tracing::info!("Pre-existing provider column in tool_executions — safely advancing migration version to 33");
+                        conn.execute_batch(
+                            "CREATE TABLE IF NOT EXISTS phantom_events (
+                                id TEXT PRIMARY KEY, session_id TEXT NOT NULL DEFAULT '', provider TEXT, model TEXT,
+                                detected_at INTEGER NOT NULL DEFAULT (strftime('%s', 'now')), resolved INTEGER NOT NULL DEFAULT 0,
+                                retry_count INTEGER NOT NULL DEFAULT 0, tools_after_retry INTEGER NOT NULL DEFAULT 0
+                            );
+                            CREATE INDEX IF NOT EXISTS idx_phantom_events_detected_at ON phantom_events(detected_at);
+                            CREATE INDEX IF NOT EXISTS idx_phantom_events_session ON phantom_events(session_id);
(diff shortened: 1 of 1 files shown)


adolfousier:
## Audit + merge notes

Merged, with follow-up fixes on top. The issue #937 is real, but the original fix had a gap:

**Gap in this PR as submitted:** the heal gated on `user_version < 33` and checked only the `provider` column. Migration 33 adds **three** columns (`provider`, `model`, `duration_ms`), and `tool_execution.rs` INSERTs with all three. A DB holding `provider` but missing `model`/`duration_ms` (possible partial state) would get stamped to v33 with the migration skipped — trading a migration-time crash for a runtime SQL crash on every tool-execution write.

**What was fixed on top (local main, not yet pushed):**
- Gate narrowed to `user_version == 32` — fires only when migration 33 is next; can never skip migrations 1–32 in pathological states
- `heal_analytics_migration_33()` checks all three columns via `PRAGMA table_info`, ALTER-adds whichever are missing, ensures analytics tables + indexes exist, then stamps v33
- Extracted `build_migrations()` for testability
- Regression tests in `src/tests/db_migration_33_heal_test.rs`: #937 scenario (provider pre-exists), partial columns, fresh-migration guard, double-run idempotency

**Validation:** fmt clean, clippy zero warnings, 5935 tests passed / 0 failed.

Thanks for the report and the initial fix @moneyacademyKE — the diagnosis was right, the heal just needed to cover all three columns.

Issue #1059, opened 2026-08-14
Enable forcing usage of max_completion_tokens


bootrecords:
## Problem
I want to use models from [Scaleway's Generative APIs](https://www.scaleway.com/en/generative-apis/) but when setting up Scaleway as a custom OpenAI-compatible provider (which it generally is), I still cannot obtain completions from the models as any request is answered with a `{"status":400,"error":"BAD REQUEST","message":"payload validation: max_completion_tokens is limited to 16384 for <model>"}`, regardless of the `max_tokens` setting in the `agents` setting in my `config.toml`.

### Steps to reproduce
(tested on opencrabs 0.3.79 and 0.3.80):

1. Use an org-based URL and API key from Scaleway as basis for a custom OpenAI-compatible provider
2. Select any model, e.g. `"qwen3.5-397b-a17b"`
3. Prompt anything

#### Expected behaviour
The prompt completes and receives an inference response

#### Actual behaviour
Opencrabs reports an "Error: Provider returned HTTP 400. Retrying won't help - the request is invalid for this provider. [...]" - Also the actual error response shows briefly in the tui, then disappears.

(I would have appended a log, but somehow the upload fails every time I try; it doesn't contain a lot more information at any rate)

## Where to change
`max_completion_tokens` is already handled in the code, but only for select model names (as defined in `uses_max_completion_tokens`), but the Scaleway API appears to require its usage for all models it provides - not only the ones from OpenAI with according names.

adolfousier:
Thanks — the reproduction was precise enough to find the real cause, and your PR working confirms it.

You're right that `uses_max_completion_tokens` is a name allowlist, so a Scaleway-hosted `qwen3.5-397b-a17b` never matches it.

But OpenCrabs already recovers from this without any configuration. On a 400 it swaps `max_tokens` ↔ `max_completion_tokens` and retries once — `src/brain/provider/custom_openai_compatible.rs:3372` and again at `:3572`:

```rust
if is_token_field_mismatch(&e.to_string()) {
    tracing::warn!("Token field mismatch for model {}, retrying with swapped fields", model);
    openai_request.swap_token_fields();
```

It missed your case by one word. The matcher requires the error text to contain `"unsupported"`:

```rust
pub(crate) fn is_token_field_mismatch(msg: &str) -> bool {
    let m = msg.to_lowercase();
    (m.contains("max_tokens") || m.contains("max_completion_tokens")) && m.contains("unsupported")
}
```

[...]

PR #1060, opened 2026-08-14
Force completion tokens config

bootrecords:
Possible solution for #1059 

Adds a config key to force all models on a provider to use `max_completion_tokens` instead of `max_tokens`.
Adds tests as well and updates documentation with an example

Also tested to success against the Scaleway API.

### DIFF
--- a/src/brain/provider/custom_openai_compatible.rs
+++ b/src/brain/provider/custom_openai_compatible.rs
@@ -2206,6 +2206,8 @@ pub struct OpenAIProvider {
     /// via Arc so clones of this provider (and the FallbackProvider that
     /// wraps it) read the same buffer.
     retry_notices: Arc<std::sync::Mutex<Vec<(u32, u32, String)>>>,
+    /// Force use of max_completion_tokens field instead of max_tokens
+    use_max_completion_tokens: Option<bool>,
 }
 
 impl OpenAIProvider {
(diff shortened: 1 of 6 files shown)


adolfousier:
Thanks for the contribution @bootrecords. The fix landed in `8e8212d8`.

`is_token_field_mismatch` now requires the field name **plus** any of several rejection signals, instead of the single word `"unsupported"`:

```rust
if !(m.contains("max_tokens") || m.contains("max_completion_tokens")) {
    return false;
}
["unsupported", "not supported", "limited to", "use max_completion_tokens", "instead"]
    .iter()
    .any(|sig| m.contains(sig))
```

So the swap-and-retry that already existed now fires on your Scaleway response, and on any other provider that words the same rejection differently.

**On swapping when the message is a cap complaint**, since `"limited to"` is genuinely ambiguous between "wrong field" and "value too high": the retry re-sends the same value under the other field name, so a real over-cap still fails and surfaces its own error. The swap cannot mask a genuine limit problem, it just costs one extra request in that case.

Four tests in `src/tests/openai_provider_test.rs`: your exact Scaleway wording, an explicit use-the-other-field instruction, and both halves of the guard — a rejection signal without the field name (`temperature is limited to 2.0`) and the field name with no sign of rejection (`request accepted with max_tokens=4096`). Neither should trigger a swap, and neither does.

Full suite green: 6658 passed, 0 failed.

@bootrecords this makes the config key in #1060 unnecessary for your case, so I would suggest closing that PR rather than reworking it — the diagnosis in it was what led here, and that was the hard part. If you hit an endpoint that rejects the field with *no* usable text in the response, reopen the config-override idea and it becomes worth having as an escape hatch. Closing this PR if issue persist with your case please feel free to reopen or submit new one.

I think that should fix it for good.

PR #1118, opened 2026-08-19
docs(readme): fix broken star history chart

Dessalines39394:
The star history chart in the README is currently broken because the upstream GitHub stargazer API is restricted, so the image no longer renders.

This change points the chart at a working data source so it displays correctly again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4495,7 +4495,7 @@ cargo test
 
 ## Star History Chart
 
-[![Star History Chart](https://api.star-history.com/svg?repos=adolfousier/opencrabs&type=date&legend=top-left)](https://www.star-history.com/#adolfousier/opencrabs)
+[![Star History Chart](https://star-history.dera.page/svg?repos=adolfousier/opencrabs&type=date&legend=top-left)](https://star-history.dera.page/#adolfousier/opencrabs&type=date)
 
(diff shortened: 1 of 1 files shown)


adolfousier:
Thanks for taking the time, but closing this one — the chart it fixes is not broken.

Checked both URLs directly:

```
api.star-history.com    → 200  image/svg+xml  60,125 bytes
star-history.dera.page  → 200  image/svg+xml  65,097 bytes
```

The current one serves a valid SVG, so this is not repairing a break — it repoints a README image to a different host. That image loads for everyone who opens the repository page, so where it is served from is a decision worth making deliberately rather than as a fix, and we would rather stay on the well-known service.

If you were genuinely seeing a broken chart, say what you saw and from where — a region-specific failure would be worth knowing about, and I would reopen this.

Separately, and prompted by this PR: we have added the Trendshift daily and weekly badges to the README in `a37fca51`, alongside the existing badges. Both URLs were fetched and confirmed to return real SVGs before committing. So the README did get some attention out of this, just not this change.

The star history chart stays as it is — it works, and it shows growth over time, which the badges do not.

Dessalines39394:
Are you a bot? This SVG is a big error message, you can't tell me you don't see it

Issue #1203, opened 2026-08-25
Tool: session_notify — agent-callable cross-session push with mechanical sender-signature injection


leshchenko1979:
### Summary

Expose the existing internal push path `deliver_to_session()` (`src/brain/agent/service/session_routes.rs`) as an agent-callable tool **`session_notify`**, enabling a compile-feedback loop: a compiling agent pushes build errors back to the editing-agent session whose commit broke the build (orchestrator/"Grok Bot" pattern: compile → attribute → push homework).

### Proposed API

| Arg | Type | Description |
|---|---|---|
| `target_session` | Uuid | Recipient session (claimed in `SESSION_ROUTES`) |
| `message` | String | Payload ("homework") |

Behavior: wrap as `QueuedUserMessage::system(...)`, call `deliver_to_session(target_session, msg)`. Delivery drains at the next tool-loop boundary, or starts a fresh turn if the target is idle — this IS the "do your homework" wake-up. If target not claimed → return error to caller (agent falls back to e.g. a PR comment).

Implementation shape: mirror `send_input.rs`; working usage of the same push path already exists at `push_result()` (`src/brain/tools/subagent/spawn.rs:83`).

### Mechanical signature injection

Problem: `QueuedUserMessage` has only `{context_text, display_text}` (`src/brain/agent/service/types.rs:193`). The recipient cannot know who sent the message and cannot reply.

Fix at the TOOL layer — sender identity never depends on model cooperation:

1. Add `from_session: Option<Uuid>` to `QueuedUserMessage`.
2. `session_notify` fills it with the **calling** session's id from execution context (the same id registered in `SESSION_ROUTES`) — mechanical injection, model cannot forget or spoof it via message text.
3. On enqueue/render prepend one canonical header line: `[session-notify from=<uuid>]`.

Terminology contract (kept minimal for LLM efficiency — three tokens):

- **session_notify** — the only verb (tool name).
- **from=<uuid>** — identity token inside the header line.
- **Session-Id:** git trailer — discovery fallback: compiler maps broken file → blame → trailer → `target_session`.

### Reply path

Symmetric by construction: recipient reads `from=<uuid>` and calls `session_notify(target_session = from)`. Routing lands on whichever surface/topic the *receiving* session owns — no reverse-routing state needed.

### Forum topics

[...]

Issue #1200, opened 2026-08-25
Forum groups: background-task & sub-agent result pushes use chat-wide latest topic instead of the session's thread_id


leshchenko1979:
### Problem

In Telegram forum groups, **background-task completion pushes and sub-agent result pushes land in whichever topic saw the most recent message — not the topic (session) that started the work.**

Sessions are already per-topic (#215): each topic resolves to its own session, and `TelegramState::register_session_chat(session_id, chat_id, topic_id)` records the topic. But the delivery path for detached work ignores it.

### Root cause

Both push types funnel through the same Telegram enqueue callback, `build_enqueue_callback`:

- `src/channels/telegram/resume.rs:70` (background tasks via `session_routes`, #940; sub-agent results via `deliver_to_session`, #1036):

```rust
let thread_id = super::send::latest_thread_id_for_chat(chat_id).await;
```

- `latest_thread_id_for_chat` (`src/channels/telegram/send.rs:39`) returns the `thread_id` of the **most recent message in the entire chat** from the channel-messages DB.

So in a forum group the resumed turn is routed to "whatever topic was last active", which diverges from the originating topic as soon as any other topic has traffic while the task runs.

The correct lookup already exists one layer up: `TelegramState::session_topic(session_id)` (`src/channels/telegram/state.rs:500`) — added for approval-callback routing (#247, #249) — but the bg-resume path never consults it.

### Repro

1. Forum group with topics A and B.
2. In topic A, ask the bot to run a long command detached (or spawn a sub-agent).
3. While it runs, send any message in topic B.
4. Task finishes → result push appears in **topic B**.

### Proposed fix

In `build_enqueue_callback`, resolve the destination from the session first:

```rust
let thread_id = match state.session_topic(session_id).await {
    Some(t) => Some(ThreadId(MessageId(t))), // the topic that owns the session
    None => super::send::latest_thread_id_for_chat(chat_id).await, // fallback
};
```

One-line-ish change at a single choke point; both background-task and sub-agent pushes are covered by it because they share the callback.

### Edge cases

| Case | Behavior |
|------|----------|
| DM / non-forum group | Unchanged (`session_topic` = None → fallback → None) |
| Bot restart before task finishes | `session_topic` map is in-memory and empty until the next inbound message re-registers → falls back to current chat-wide-latest lookup (today's behavior, still possibly wrong topic) |
[...]

Issue #1206, opened 2026-08-25
Background-task & subagent results silently dropped for sessions revived by startup crash-recovery until next inbound message


leshchenko1979:
## Environment

OpenCrabs v0.3.83, Telegram channel daemon (`opencrabs-ops`, forum-topic chat), Linux, headless.

## Summary

When a session is resumed by **startup crash-recovery** (not by an inbound channel message), neither background-task completions nor sub-agent results can reach its channel. They fall into a silent void: no resume, no queue log, no warning of any kind. The session stays unclaimed until its channel receives the *next inbound user message*, because `claim_for_channel` is called only in ingress handlers.

## Evidence (2026-08-25, single daemon, ~1 hour)

| Time (UTC) | Event | Outcome |
|---|---|---|
| 17:17:50 | daemon boot; a session is revived by startup crash-recovery (last real inbound: 17:05) | session executes on boot surface, never claimed |
| 17:40:35 | that session's bg task finishes mid-turn | **silent drop** |
| 17:40:24 | control: another session with post-boot inbound traffic, bg task finishes while idle | ✅ resumed in ~1s, result delivered to its own forum topic |
| 17:48:03 | first session's second bg task finishes (`exit=0`, elapsed 301s) | **silent drop** — zero log lines between completion and hours later |

The control case proves the resume path itself works; the only variable is whether the session was claimed after boot.

## Root cause

1. `src/brain/agent/service/session_routes.rs` — `SESSION_ROUTES` is populated **only** via `claim_for_channel` from ingress handlers (`src/channels/telegram/handler.rs`). Crash-recovered turns enter through `resume_session` (documented "Called from ui.rs on startup") and never claim.
2. `src/brain/agent/service/background_tasks.rs` (push path): unclaimed session → `resolve_route` falls back to the *executing* surface's enqueue callback — on a headless daemon that's the boot surface, i.e. nowhere.
3. `src/brain/tools/subagent/spawn.rs::push_result`: same failure mode via `deliver_to_session`'s fallback.
4. The codebase already has the right machinery — `restart_recovery::deliver_or_park` ("Nothing is dropped on either branch") — but it was only used by startup-recovery reports (#1037), not by these two consumers.

Note this is distinct from #1200 (wrong topic selected when routing happens); here routing never happens at all.

## Proposed fix

Route both consumers through `deliver_or_park` when no route claims the session:

- fast path unchanged when claimed;
- parked otherwise;
[...]

PR #1207, opened 2026-08-25
feat(tools): session_notify — cross-session push with mechanical sender attribution (#1203)

leshchenko1979:
## Summary

Adds **`session_notify`**, a new agent-facing tool for cross-session message push **within a single OpenCrabs process**. Any session can deliver a queued system message to any other live session by UUID; the receiving session drains it at its next tool-loop boundary, or immediately (fresh turn) if idle.

**Motivation.** Multi-agent workflows need an orchestrator/compiler session to hand results back to worker/editor sessions — the canonical loop being "compile the whole binary → push compile errors to each editing agent's session → editor fixes its own code". Today the only push path is sub-agent completion (`push_result()` in `spawn.rs`), which is tied to a parent-child relationship and only fires once, when the child finishes. `session_notify` generalizes delivery to arbitrary session pairs, in both directions.

Refs #1203 (part 1 of 2 — the session-discovery part lands separately) · Related: #1200 (forum-topic routing of background pushes — see caveats).

## Design

- Thin wrapper over the existing internal primitive `deliver_to_session(session_id, QueuedUserMessage)` (`src/brain/agent/service/session_routes.rs`) — the exact path sub-agent results already ride. No new transport, registry, or queue.
- Parameters: `{ target_session: Uuid, message: String }`.
- **Mechanical sender attribution.** The calling session's UUID is taken from `ToolExecutionContext.session_id` (execution context), never from model-generated text. On enqueue the tool prepends a `[session-notify from=<uuid>]` header to the payload. A receiving agent can always reply later by calling `session_notify` back with that UUID — reply routing comes free. The calling model cannot forge or omit the sender identity.
- Payload is built as a system `QueuedUserMessage`, so transcripts show a compact tagged entry instead of raw scaffolding.
- Delivery failure (target not registered: unknown id, different instance/profile, or not seen since boot) returns an error hinting at `a2a_send` for cross-instance targets.

Ontology kept deliberately tiny for LLM efficiency — three tokens total: `session_notify`, `from=<uuid>`, `Session-Id:` trailer.

## Changes

| File | Change |
|---|---|
| `src/brain/tools/subagent/notify.rs` | new tool (~130 lines), mirrors sibling `send_input.rs` structure |
| `src/brain/tools/subagent/mod.rs` | module wiring |
| `src/brain/tools/tool_setup.rs` | registration (agent tools 8→9) |
| `src/brain/tools/catalog.rs` | discovery-catalog entry + category arm |

No exhaustive-list test changes needed; assertions on these surfaces are additive.

## Verification

- **Compiles green:** built in CI (fork quick-build run 32877632142 @ merge sha `10c73644`; content of these 4 files on this branch proven byte-identical via empty `git diff <merge> <branch> -- <files>`).
- **Live end-to-end smoke test passed** in production: registry lookup → enqueue → mechanical header injection → drain at next tool boundary; received payload carried the intact `[session-notify from=…]` header.
- **Caller self-id already solved:** `session_context {operation:"summary"}` prints the session's own UUID (first line) — agents can sign commits with `Session-Id:` trailers without any new plumbing.
- **Lint gate:** baseline-diffed against upstream main — delta = 3 strict advisories on new lines only, all matching existing sibling-file house patterns (import style ≡ `send_input.rs:11`, qualified path style ≡ `spawn.rs:90`). Zero errors.

## Known limitations / future work

- **Same-process only.** `SESSION_ROUTES` is an in-process registry; pushing across instances/profiles stays on `a2a_send`.
- **Restart edge.** Routes are re-claimed by a session's first message after daemon boot; a push to a not-yet-seen session falls back to the booting surface instead of that session's forum topic.
- **Follow-up (tracked in #1203):** structured session discovery in `session_search` — print ids (short form) + last-active timestamps, sort descending, add a machine-readable `query` op — so agents can find target UUIDs without parsing commit trailers.

### DIFF
--- a/src/brain/tools/catalog.rs
+++ b/src/brain/tools/catalog.rs
@@ -135,6 +135,7 @@ pub const EXTENDED_TOOL_INVENTORY: &[(&str, &[&str])] = &[
             "send_input",
             "close_agent",
             "resume_agent",
+            "session_notify",
             "team_create",
             "team_delete",
             "team_broadcast",
@@ -277,6 +278,7 @@ pub fn tool_category(name: &str) -> &'static str {
             || n.starts_with("send_input")
             || n.starts_with("close_agent")
             || n.starts_with("resume_agent")
+            || n.starts_with("session_notify")
             || n.starts_with("team_") =>
         {
             "agents"
--- a/src/brain/tools/subagent/mod.rs
+++ b/src/brain/tools/subagent/mod.rs
@@ -10,6 +10,7 @@
 pub mod agent_type;
 mod close;
 pub mod manager;
+mod notify;
 pub mod reconcile;
 mod resume;
 mod send_input;
@@ -22,6 +23,7 @@ pub(crate) mod worktree;
 pub use agent_type::{build_child_registry, map_deprecated_agent_type};
 pub use close::CloseAgentTool;
 pub use manager::{SubAgent, SubAgentManager, SubAgentState};
+pub use notify::SessionNotifyTool;
 pub use resume::ResumeAgentTool;
 pub use send_input::SendInputTool;
 pub use spawn::SpawnAgentTool;
--- a/src/brain/tools/subagent/notify.rs
+++ b/src/brain/tools/subagent/notify.rs
@@ -74,7 +74,7 @@ impl Tool for SessionNotifyTool {
             .and_then(|v| v.as_str())
             .ok_or_else(|| ToolError::InvalidInput("'message' is required".into()))?;
         if message.trim().is_empty() {
-            return Ok(ToolResult::error("Refusing to send an empty message"));
+            return Ok(ToolResult::error("Refusing to send an empty message".to_string()));
         }
 
         // Mechanical sender signature — from execution context, not model text.
--- a/src/brain/tools/subagent/notify.rs
+++ b/src/brain/tools/subagent/notify.rs
@@ -1,0 +1,121 @@
+//! session_notify tool — push a message into another live session's queue.
+//!
+//! Thin agent-facing wrapper over `session_routes::deliver_to_session`, so an
+//! orchestrator (e.g. a compiling agent) can hand work back to the sessions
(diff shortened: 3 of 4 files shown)


adolfousier:
Merged locally at `00b68a2c` and landing with the next push. Thank you — the design is right: a thin wrapper over `deliver_to_session` with no new transport, and taking the sender id from `ToolExecutionContext` rather than model text is the part that makes reply routing trustworthy.

Two things needed adjusting on top, both my doing rather than yours.

**1. `deliver_to_session` grew a third outcome between your branch and main.** I landed #1206 yesterday, which added *parking*: a session whose channel has not claimed it since a restart now holds the message instead of falling back to whatever surface booted. Against the old two-state bool that reads as failure, so `session_notify` told the caller

> No live route for session … — it has not messaged since boot

for a message it had in fact queued and which would be delivered as soon as the channel bound the session. Exactly the opposite of what happened.

Rather than patch the string, I made the primitive return the three outcomes it actually has (`9c0e4b81`):

```rust
pub enum Delivery { Delivered, Parked, NoRoute }
```

`session_notify` now reports a park as queued (`28bd3391`). The sub-agent path benefited too: it was quieter but no better, logging nothing at all for either a park or a genuine drop.

**2. Tests moved to `src/tests/`.** House rule here is that no inline `#[cfg(test)] mod tests` blocks live in source files; every test is a file under `src/tests/` registered in `mod.rs`. Both of your assertions are kept verbatim in `src/tests/session_notify_test.rs`, plus the delivery-reporting cases the bool-era tool could not express (`c650d425`).

Everything else went in as written. Capabilities and `requires_approval` match the sibling `send_input` exactly, which is what I checked them against.

**Verification:** `cargo clippy --locked --lib --bins --tests --all-features -- -D warnings` clean, `7069 passed; 0 failed`, `cargo fmt --check` clean. Note that main was clippy-red when you opened this (#1212, now fixed in `b9b435e1`), so your red checks were never about your code.

Issue #1202, opened 2026-08-25
Telegram rich: render untagged fenced Mermaid blocks as diagrams (content-based classification)


leshchenko1979:
## Problem

The Telegram rich renderer (src/channels/telegram/rich/mermaid.rs) only treats a fenced block as Mermaid when the fence is explicitly tagged ```mermaid. LLM output very often emits bare fences:

````
```
graph TD
    A --> B
```
````

These currently render as a plain code block (or fall through to the HTML fallback as Code{lang: None}) instead of being rendered as a diagram. The model was clearly speaking Mermaid; only the tag is missing.

## Proposed behavior

When a fenced block has **no language qualifier**, classify it by content: if the first non-blank, non-%% line starts a known Mermaid opener — flowchart / graph / sequenceDiagram / classDiagram / stateDiagram / erDiagram / journey / gantt / pie / quadrantChart / mindmap / timeline / gitGraph / xychart / requirementDiagram / C4Context etc. — treat it exactly like a tagged mermaid fence and render it as an image.

Guard against false positives: a bare `graph` must be followed by a direction word (TD/TB/BT/LR/RL) so Graphviz DOT sources (`graph G {`) stay plain code blocks.

Explicitly tagged ```mermaid behavior is unchanged.

## Implementation sketch

1. New classifier looks_like_mermaid_source(source: &str) -> bool in src/channels/telegram/rich/mermaid.rs — checks first non-blank/non-comment line against the opener table (with the graph+direction guard).
2. Wire it into all three render paths:
   - has_mermaid_fence() — async gate (now also returns true when an untagged closed fence classifies as mermaid)
   - find_mermaid_fences() — capture gate at open time (rest.is_empty() && looks_like_mermaid_source(body))
   - resolve_block() — match guard widened from lang == Some("mermaid") to also accept lang == None && looks_like_mermaid_source(&text)

## Patch

Full implementation + tests (+127/−19, 2 files), on my fork:

- Branch: https://github.com/leshchenko1979/opencrabs/tree/fix/bg-push-session-topic
- Patch link: https://github.com/leshchenko1979/opencrabs/commit/3c9b92b0.patch

Tests included: detection positives per opener family, rejections (SQL, DOT digraph, bare `graph`, bare unclassifiable fence), byte-range extraction on an untagged fence, mixed tagged/untagged ordering, unclosed-fence pass-through.

leshchenko1979:
[...]

Issue #1044, opened 2026-08-14
feat(telegram): render mermaid diagrams from LLM output as images alongside rich messages


adolfousier:
## Summary

LLMs routinely emit ```mermaid fences (verified: Qwen3.8-Max via modelscope outputs them natively in markdown replies). Telegram `sendRichMessage` rich HTML supports native `<img>` blocks, so OpenCrabs can render those fences as inline diagram images inside rich messages — no hosting, no companion `sendPhoto`, diagram and text in one message.

## Prototype evidence (verified live 2026-08-14)

| Check | Result |
|---|---|
| Model output format | standard ```mermaid fenced code block (Qwen-Ambassador/Qwen3.8-Max, verbatim capture) |
| mermaid.ink encoding | **base64url** (RFC 4648 §5); standard base64 → 404 |
| Valid diagram | HTTP 200, image (jpeg/png varies), renders via `<img src>` inside a rich message (prototype message_id 1047) |
| Broken diagram | HTTP 400 + legible text/plain: full parse error (line, source, caret, expected tokens); garbage input → "Unknown diagram error" |
| Broken `<img src>` embedded | **sendRichMessage itself fails: 400 `RICH_MESSAGE_PHOTO_NO_MEDIA_FOUND`** — one broken diagram kills the entire message |

The last row decides the degradation design: mermaid.ink's own signaling is insufficient, the pipeline must pre-validate and emit an explicit legible failure block. The failure path prototype was delivered to the group thread: heading + explanation + blockquote carrying mermaid.ink's actual parse error + original broken source in a code block.

## Design

Pipeline (insertion points verified in-tree):

1. **Detection** — `src/channels/telegram/rich/parse.rs:29-42` already consumes fences and extracts the language (`fence_lang`, parse.rs:202) into `Block::Code { lang, text }` (ast.rs:24). Mermaid detection is `lang == Some("mermaid")`. Zero parser changes.
2. **Pre-validation** (new module `src/channels/telegram/rich/mermaid.rs`) — for each mermaid fence: base64url-encode source → GET `https://mermaid.ink/img/<b64>` with bounded timeout → expect 200 + `image/*`. Success: embed the URL. Failure: capture mermaid.ink's error text.
[...]

PR #1208, opened 2026-08-25
telegram: render untagged Mermaid fences; surface diagram-bearing intermediates

leshchenko1979:
Refs #1202 (problem statement + patch links there).

## Problem

LLM-generated messages frequently wrap Mermaid in a bare ``` fence without the `mermaid` info string. The rich renderer only treated fences tagged `mermaid` as diagrams (#1044), so these rendered as gray code boxes. Separately, assistant narration emitted *before* a tool call ("intermediates") is folded into the expandable flow blockquote as escaped plain text — so even correctly tagged diagrams vanished behind tap-to-expand as raw source.

## Changes

**Commit 1 — content-based untagged classification**
- new `looks_like_mermaid_source()`: first non-blank, non-`%%` line must start a known opener (`flowchart`, `graph`, `sequenceDiagram`, `classDiagram`, `stateDiagram`, `erDiagram`, `gantt`, `pie`, `mindmap`, `timeline`, `gitGraph`, …)
- a bare `graph` additionally requires a direction token (`TD/TB/BT/LR/RL`), so Graphviz DOT sources stay plain code
- wired into all three render paths: fence-detection gate, markdown+media path, and the HTML fallback (`Code{lang: None}`)

**Commit 2 — intermediate surfacing**
- the deliverable-rich gate for pre-tool-call narration now also accepts any closed mermaid fence (tagged or content-classified), so diagram-bearing intermediates surface as their own rendered bubble instead of folding into the collapsed flow log
- surfaced sends route through the mermaid-resolving sender; the HTML fallback resolves fences too
- non-fence rich reports keep byte-identical behavior

## Tests

New cases in `telegram_mermaid_test.rs` / `telegram_flow_chrome_test.rs`: detection positives per opener, rejections (SQL, DOT `graph G {`, bare unclassifiable fence), byte-range extraction on an untagged fence, mixed tagged+untagged ordering, unclosed-fence pass-through, surfacing-gate coverage.

## Verification

Deployed on our fork build and verified live in production against Telegram's renderer: untagged fences render as diagrams, and diagram-bearing intermediates surface as standalone rendered bubbles.

### DIFF
--- a/src/channels/telegram/intermediates.rs
+++ b/src/channels/telegram/intermediates.rs
@@ -59,7 +59,9 @@ pub(crate) fn build_last_intermediate_with_footer(
 }
 
 /// Send a structured intermediate segment as a native rich message, returning
-/// its id for tracking. Returns `None` when the text carries no rich structure
+/// its id for tracking. Mermaid-aware (#1044/#1202): fences resolve to a media
+/// array; with no fence this is byte-identical to `send_rich_markdown_id`.
+/// Returns `None` when the text carries no rich structure
 /// or the rich API rejects it — the caller then falls back to the HTML path.
 pub(crate) async fn try_send_intermediate_rich(
     bot: &Bot,
@@ -70,7 +72,10 @@ pub(crate) async fn try_send_intermediate_rich(
     if !super::rich::should_send_native_rich(text) {
         return None;
     }
-    match super::rich::api::send_rich_markdown_id(
+    // Mermaid-aware sender (#1044/#1202): resolves fences into the rich
+    // markdown media array; byte-identical to send_rich_markdown_id when no
+    // fence is present, so non-diagram reports are unaffected.
+    match super::rich::send_rich_with_mermaid_id(
         bot.api_url().as_str(),
         bot.token(),
         chat_id.0,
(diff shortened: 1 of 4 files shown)


adolfousier:
Merged locally at `79ae62925` and landing with the next push. Thank you — content-based classification is the right call for this, and surfacing diagram-bearing intermediates rather than folding them into the collapsed log fixes a genuinely annoying behaviour. The `graph` + direction-token guard against DOT is a nice touch.

**One correctness fix on top** (`5d0a5a0e7`).

Both line scanners decide "is this fence untagged?" from the info string of the **closing** fence:

```rust
if tagged_mermaid || (rest.is_empty() && looks_like_mermaid_source(body)) {
```

At that point `rest` is whatever follows ``` on the closing line, which is bare on essentially every block — so `rest.is_empty()` is true regardless of what the opening fence said, and content classification ran on tagged fences too. Verified before changing anything:

| input | before | after |
| --- | --- | --- |
| ` ```dot ` + `graph TD …` | classified as mermaid | plain code |
| ` ```text ` + `sequenceDiagram …` | classified as mermaid | plain code |
| ` ``` ` (bare) + `graph TD …` | classified | classified |
| ` ```mermaid ` | classified | classified |

This contradicted your own doc comment ("untagged fences, whose info string is empty") and disagreed with `resolve_block`, which already gets it right with `lang.is_none()`. So the three render paths were applying two different rules. Both scanners now remember the opening info string.

Your DOT rejection test passed for a different reason than it appears to: `graph G {` is rejected because it lacks a direction token, not because the fence was tagged. With the tag ignored, a ` ```dot ` block containing `graph TD` would still have rendered as a diagram.

New cases in `src/tests/telegram_mermaid_test.rs` pin both directions (`38e1a4616`). All of your existing cases pass unchanged.

**Verification:** clippy clean with the full CI flags, `7078 passed; 0 failed`, fmt clean. Your red checks were main's fault, not yours — #1212, fixed in `b9b435e1`.

PR #1113, opened 2026-08-19
feat(telegram): optional userbot channel via grammers (QR/code/2FA login, read-only user plane)

moneyacademyKE:
## Title
feat(telegram): optional userbot channel via grammers (QR/code/2FA login, read-only user plane)

### What
Adds an optional `telegram-userbot` cargo feature implementing a second Telegram
plane alongside the existing Bot API channel, using the pure-Rust MTProto stack
(`grammers-client`, `grammers-session`, `grammers-mtsender`, `grammers-crypto`
0.10). The user plane is **read-only**: inbound updates from the personal
account are converted to the existing Bot API message shape and fed to the same
agent dispatch path (`handle_message`); all outbound replies still exit via the
bot plane. There is **no outbound-as-user path** in this PR by design.

### Why
Bot API tokens cannot read the personal account's dialogs (groups the bot was
never added to, DMs, channels). A user session sees everything the user sees.
Keeping the plane read-only avoids acting as the user — the risky half, both
technically and per Telegram ToS.

### How it works
- **Config**: `[channels.telegram.userbot]` with `enabled`, `api_id`, `api_hash`,
  `phone`, `session_path`, `allowlist` (chat IDs permitted to reach the agent;
  empty = nothing forwarded). Secrets merge from `keys.toml` with a sentinel
  (`__KEYS__`) so config.toml never carries real creds.
- **Login**: `opencrabs channel userbot-login` — QR-first (`auth.exportLoginToken`,
  terminal-rendered via `qrcode` crate, auto DC-migration via `importLoginToken`),
  code fallback (`request_login_code`/`sign_in`), and 2FA SRP
  (`account.getPassword` + `calculate_2fa` + `auth.CheckPassword`) replicating
  grammers' private `check_password` with public API only.
- **Watch loop**: `SenderPool` update stream → grammers `Message` → Bot API JSON
  bridge (`bot_api_dialog_id()` peer mapping) → allowlist gate → existing
  `handle_message` dispatch with reaction/resume callbacks wired like the bot plane.
- **Manager**: `reconcile_telegram_userbot` start/stop/noop mapping, restart on
  dead handle, same lifecycle as other channels.

### Session file handling
`userbot.session` (SQLite via `SqliteSession`) is the full account credential and
is treated like bot tokens: stored under the OpenCrabs home, never in the repo,
documented in README section added here.

### ToS / ban-risk note
Automated user accounts are against Telegram ToS when visible to other users
[...]

PR #1209, opened 2026-08-25
feat(telegram): add receive-only userbot ingestion

### LINKED ISSUE #1112: feat: Telegram userbot — read-only user plane via grammers (QR/code/2FA login)
## Problem
The Bot API plane cannot see the personal account's dialogs: groups the bot was never added to, DMs, channels. A user-session (MTProto) plane sees everything the user sees, but nothing in OpenCrabs speaks MTProto.

## Proposal
Optional `telegram-userbot` cargo feature adding a second, **read-only** Telegram plane alongside the existing Bot API channel, built on pure-Rust grammers 0.10:

- **Login**: `opencrabs channel userbot-login` — QR-first (`auth.exportLoginToken`, terminal-rendered, auto DC-migration via `importLoginToken`), code fallback, 2FA SRP. Session persists as SQLite (`SqliteSession`) and is treated like bot tokens (never in repo, OpenCrabs home).
- **Watch loop**: allowlist-gated update stream; own sends and bot traffic skipped; messages converted to the existing Bot API envelope (`bot_api_dialog_id` peer mapping) and fed to the same `handle_message` dispatch — replies exit as the bot.
- **No outbound-as-user path** in v1, by design.

## Why read-only
Acting as the user is the risky half — technically (wrong-chat sends) and per Telegram ToS (automated user accounts are spam-adjacent). Read-only + allowlist is the lowest-profile shape; outbound-as-user should only ever land as a separately-reviewed allowlist feature.

## Evidence
Full spike completed before integration: QR login + DC migration + 2FA SRP verified against a live account, session persistence across processes verified, live update receive + message extraction verified. Map with pinned grammers 0.10 API shapes (docs are two majors stale): grammers 0.10 replaced `Client::connect`/`Config` with `SenderPool` construction; QR and SRP require raw-TL invocation with public session-trait persistence.

moneyacademyKE:
## Summary

Adds an optional, receive-only Telegram MTProto capture plane alongside the existing Bot API channel.

This replacement is intentionally scoped to one trust boundary:

- local QR / phone-code / hidden 2FA login from the CLI
- local JSON session persistence with atomic writes and `0600` permissions
- live inbound text capture from explicitly allowed chats
- pre-conversion, pre-storage allowlist gate; empty means dry mode
- independent ChannelManager lifecycle; no Bot API token required
- passive storage under `telegram-userbot`, retrieved explicitly through `channel_search`
- no automatic LLM invocation and no automatic reply

## Explicit non-goals

- no userbot tool plane
- no send, edit, react, schedule, wait-for-reply, or send-to-phone path
- no media download
- no raw MTProto invocation registry
- no outbound permission model
- no conversational credential collection in Telegram chats

The executable source-boundary test rejects account mutation, media download, and agent dispatch in the userbot module.

## Why a replacement

#1113 began with a receive-only contract but later grew interactive reads, account mutation, media I/O, and raw MTProto access. Those are separate trust boundaries and are deliberately absent here.

## Verification

On Rust 1.98, against this branch plus the separate main-baseline repair in #1215:

- `cargo fmt --all -- --check` — pass
- `cargo clippy --locked --lib --bins --tests --all-features -- -D warnings` — pass
- `RUST_MIN_STACK=16777216 cargo test --locked --all-features` — 7,076 passed, 0 failed
- focused `telegram_userbot` tests — 15 passed, 0 failed

#1215 contains only the 21 pre-existing clippy fixes from current `main`; none of those paths are present in this PR.

Fixes #1112


moneyacademyKE:
Verification update: the branch remains unchanged at `eb169be7`. The 21 failures in the existing Lint check are reproduced on current `main` and fixed separately in #1215. On a detached verification worktree combining this exact head with #1215, Rust 1.98 `fmt` and the exact CI `clippy` command pass; focused userbot tests are 15/15 and the full all-features suite is 7,076 passed / 0 failed. No #1215 paths are present in this PR.

moneyacademyKE:
Rebased onto current `main` (`2132927a`), which now carries the baseline lint fixes that closed #1215 — thank you for landing those. This branch is back to containing only the receive-only userbot changes (plus one rustfmt module-ordering commit). Verified locally on the CI toolchain (1.98.0): `fmt --check`, `clippy --locked --lib --bins --tests --all-features -- -D warnings` (exit 0), userbot-focused tests 15/15, full suite 7,131 passed / 0 failed / 29 ignored. Note: the upstream baseline is now green, so the earlier Lint failures against the old base are resolved by this rebase.

Issue #1210, opened 2026-08-25
tool_search activation never reaches subagent sessions — searched tools stay uncallable in children


leshchenko1979:
# tool_search activation never reaches subagent sessions — searched tools stay uncallable in children

**Follow-up to #1025.** The fix there (`ea0f33b6`) wired `tool_search` → `activate_tools`, but only for the session whose service constructed the tool. In **subagent (child) sessions** the activation lands in the **parent's** registry under the child's session id, while the child's request builder reads the **child's** registry — so the searched tool's schema never rides on the child's requests, and the model can never call what `tool_search` just told it was callable.

#1025's closing comment said: *"Not verified live: the check is one spawn request on the model that failed — it should emit `spawn_agent` rather than narrate it."* This issue IS that live check, run 2026-08-25 with four subagent spawns across two models on two providers. **Every run failed the same way.**

## Repro

Spawn a subagent with one task: use `tool_search("telegram send message")` then call `telegram_send` exactly once to post `SMOKE4`. Lazy-tools on (default). Observed across attempts:

| # | model / provider | behavior |
|---|---|---|
| 1 | deepseek-v4-flash (opencode) | 70+ bash calls, never emitted `telegram_send` → killed |
| 2 | deepseek-v4-flash | misrouted to `http_request`; result correctly returned to parent |
| 3 | deepseek-v4-flash (resume) | `tool_search("telegram send message")` ×4 → loop-breaker fired |
| 4 | kimi-k2.7-code | byte-identical `tool_search` ×4 → loop-breaker fired |
| 5 | kimi-k2.7-code (resume, told to skip tool_search and call telegram_send directly) | still `tool_search` ×4 → loop-breaker fired |

Two models on two providers converging on the byte-identical loop (`args_len=33`) is the signature of a tool *described but not declared*: well-behaved models will not emit a call for a function absent from their tool list, so they re-search forever.

## Log evidence (2026-08-25)

Child session `f2e304df-dacb-4459-9bcf-56d9366fc54b` (kimi):

```
20:28:54  DEBUG tool_loop.rs:1711: Tool registry contains 61 tools
20:28:54  WARN  custom_openai_compatible.rs:2625: NO SYSTEM BRAIN in request!
20:29:10  DEBUG tool_loop.rs:1711: Tool registry contains 61 tools
[...]

Issue #1025, opened 2026-08-12
tool_search describes a tool but never makes it callable: spawn_agent loops until the turn is killed


adolfousier:
`tool_search` returns a tool's schema as TEXT and never adds it to the session's active set, so the schema does not ride on the next request. The model is told the tool exists, believes it is available, and has no such function to call.

Observed twice on the same task, on `zhipu/glm-5.2`:

- Run 1: 22 tool calls, 3:12. `tool_search("spawn subagent for independent review task")` returns OK, then "Spawning the subagent now" → `bash(echo "spawning subagent for issue #1022")`. Later `tool_search("spawn_agent subagent")` again, "Now spawning with the full task:" → `read_file(navigate.rs)`. Killed by the announcement-loop detector.
- Run 2: 8 tool calls, 1:07. Same shape, the model even reporting "Spawn tool active" after the search, then "Spawning now" → `bash(echo "Spawning validation subagent...")`. Killed again.

Removing the spawn from the request made the same task complete with no issues. The model never emitted a single `spawn_agent` call across 30 tool calls.

## Cause

`registry.rs:106` `activate_tools` is what puts an EXTENDED tool's schema into subsequent requests. Its only caller is `registry.rs:306`, the JIT-on-execute path, which runs when a non-core tool is **actually used**.

`tool_search.rs` never calls it. From `tool_search.rs:90`:

```
// We deliberately do NOT pre-activate all matches (#604): activation happens
// on actual use via the JIT-on-execute path, so only the tool the model really
// calls gets its structured schema injected
```

That is circular for any model that will not emit a call for a function absent from its tool list: activation requires a call, and the call requires activation. Capable models escape by calling from the text schema; weaker ones substitute something that IS active — here `bash` echoes and `read_file`.

The module's own doc comment at `tool_search.rs:17` reads "Max tools returned (and activated) per search", so activation was the intent; the call was never wired.

## Blast radius

Every search-gated capability, not just spawning. The tool's description names browsing, channel sends, sub-agents and teams, image and video generation, cron, and self-improvement as search-first. All of them depend on the model emitting a call for a function it cannot see.

It also defeats the announcement-loop detector's purpose: the detector fires correctly, but the underlying cause is not something rephrasing or nudging can fix.

## Scope

- `tool_search` activates the tools it returns, so their schemas ride on the next request
- Activation stays bounded — matches are already capped per search, and `activate_tools` has LRU eviction at `MAX_ACTIVE_EXTENDED`
[...]

PR #1214, opened 2026-08-25
fix(subagents): re-bind tool_search to each child registry so runtime-activated tools work in children

leshchenko1979:
## Problem

Tools activated at runtime via `tool_search` are never callable inside **subagent sessions**. A child that searches for a tool receives the schema back as text, then re-runs `tool_search` forever because the tool never appears in its request's tool set. Observed live (#1210): two independent models on two providers (deepseek-v4-flash/openrouter, kimi-k2.7-code/opencode) converged byte-identically on this loop — four identical `tool_search` calls each, zero calls to the searched tool — until the harness loop-breaker killed the turn.

## Root cause

The activation write and the request-build read hit **different registry instances**:

- `ToolSearchTool::new(tool_registry.clone())` binds the discovery tool to the **parent** registry Arc once at setup (`cli/tool_setup.rs:116`)
- `build_child_registry` copies the *same* `Arc<dyn Tool>` instances into a fresh child registry (`src/brain/tools/subagent/agent_type.rs`) — children stabilize at 61 tools vs parent 71
- a child's search activates matches into the captured (**parent**) registry under the *child's* session id (`src/brain/tools/tool_search.rs:115`)
- the child's request builder reads `self.tool_registry.active_tools(session_id)` = the **child** registry (`src/brain/service/helpers.rs:239`) → active set empty → schema never rides the request

#1025 added the activation plumbing plus tests, but its tests exercise a single registry in isolation — none mounts a child service whose copied search tool still points at the parent registry. #1025 was closed with an explicit "not verified live" caveat; running that live check is what surfaced this (#1210).

## Fix

`build_child_registry` now returns `Arc<ToolRegistry>` and re-binds a **fresh `ToolSearchTool`** onto the child's own registry whenever it copies one in, so every nesting level points discovery at its own registry. The three callers (`spawn.rs`, `resume.rs`, `team/create.rs`) pass the Arc straight through.

## Regression test

Extends `tool_search_activation_test.rs`: mounts a parent registry with the discovery tool bound to the parent exactly as production wires it, builds a child, executes the child's search instance, then asserts the activation landed in the **child's** active set and never touched the parent's. Signature-agnostic by design — it compiles against pre-fix code and fails there by assertion, not compile error.

## Evidence

- Live failure (#1210): 4 spawns / 2 models / 2 providers; child registry pinned at "61 tools" across every `tool_search`; loop-breaker fired twice (20:29:33Z and 20:31:54Z on 2026-08-25)
- Green build containing this fix: [leshchenko1979/opencrabs run 32903857601](https://github.com/leshchenko1979/opencrabs/actions/runs/32903857601) (merge commit `21099265`)
- Deployed: our production ops instance runs a sha256-verified binary built from this fix
- Honest caveat: the new regression test has compiled everywhere but has not yet executed in fork CI — cancel-in-progress churn on a busy fork main kept killing full runs before the test job started; verification here was modum-gated

Refs #1210
Refs #1025

### DIFF
--- a/src/brain/tools/subagent/agent_type.rs
+++ b/src/brain/tools/subagent/agent_type.rs
@@ -13,7 +13,10 @@
 //! bash, so it was write-capable). Unknown values fail closed rather than
 //! silently escalating to full write access.
 
-use crate::brain::tools::ToolRegistry;
+use crate::brain::tools::catalog;
+use crate::brain::tools::registry::ToolRegistry;
+use crate::brain::tools::tool_search::ToolSearchTool;
+use std::sync::Arc;
 
 /// Tools that sub-agents must NEVER have access to (prevents recursion / dangerous ops).
 const ALWAYS_EXCLUDED: &[&str] = &[
@@ -34,11 +37,31 @@ const ALWAYS_EXCLUDED: &[&str] = &[
 /// This is the FULL-access child baseline. When the parent granted only read
 /// access, the caller additionally applies plan_gate's
 /// `restrict_registry_to_read_only` on the returned registry.
-pub fn build_child_registry(parent: &ToolRegistry) -> ToolRegistry {
-    let child = ToolRegistry::new();
+///
+/// Returns an [`Arc`] because callers hand the registry to `AgentService`
+/// wrapped anyway — and because [`ToolSearchTool`] must be RE-BOUND to this
+/// child Arc (#1210): the parent's search-tool instance captured the parent's
+/// registry at construction. Copying that instance here made a child's
+/// `tool_search` activate schemas in the PARENT's registry under the child's
+/// session id, while the child's request builder read active tools from its
+/// OWN registry — so a discovered tool's schema never rode on the child's
+/// next request, and well-behaved models looped re-searching forever instead
+/// of emitting calls for undeclared tools (observed live on two models across
+/// two providers; see upstream #1210). Re-binding per level also keeps the
+/// fix correct for nested spawns: each generation's search tool points at
+/// its own registry.
+pub fn build_child_registry(parent: &ToolRegistry) -> Arc<ToolRegistry> {
+    let child = Arc::new(ToolRegistry::new());
 
     for name in parent.list_tools() {
         if ALWAYS_EXCLUDED.contains(&name.as_str()) {
+            continue;
+        }
+
+        if name == catalog::TOOL_SEARCH_NAME {
+            // Fresh instance bound to THIS registry, so the child's searches
+            // land where the child's requests read (#1210).
+            child.register(Arc::new(ToolSearchTool::new(child.clone())));
             continue;
         }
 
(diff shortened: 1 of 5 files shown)


adolfousier:
Merged locally at `db8d8a25` and landing with the next push. Thank you — and specifically for putting the re-binding **inside `build_child_registry`** rather than at the call site.

I had implemented this independently before seeing your PR, and I fixed it at `spawn.rs` only. That misses `resume.rs:207` and `team/create.rs:277`, so resumed sub-agents and team members would have stayed broken, and any future builder would have inherited the bug. Yours is one choke point and is also correct for nested spawns, where each generation binds its own. I took your structure and dropped mine.

**One change on top** (`63faa87a`): the binding is now `Weak` rather than `Arc`.

```rust
child.register(Arc::new(ToolSearchTool::new(&child)));  // was child.clone()
```

The registry owns the tool, so handing it a strong `Arc` back closes a `registry -> tool -> registry` cycle and the child registry is never freed. That was already true of the parent, where it costs one leak for the process lifetime and nobody notices. Per spawn it is a different matter, and the re-binding is what turns the latent case into a live one — the fix would otherwise have traded a correctness bug for a slow leak.

`ToolSearchTool` now holds `Weak<ToolRegistry>` and upgrades at call time; failure returns an explicit error rather than an empty result set, since "no tools found" is exactly the conclusion the model must not reach here.

Pinned by a test that fails under `Arc` and passes under `Weak`:

```rust
let watch = { let child = build_child_registry(&parent); Arc::downgrade(&child) };
assert!(watch.upgrade().is_none());
```

**On the tests.** Yours is the primary coverage and I kept it as-is (one call-site adjustment for the new signature). Mine is trimmed to what yours does not cover: the drop assertion above, and the pre-fix mechanism built by hand, since `build_child_registry` now always re-binds and can no longer produce it. I kept that one deliberately — it is what the report observed, and it is what #1025's tests could not see, because they exercise a single registry where parent and child are the same object.

**Verification:** clippy clean with the full CI flags, `7089 passed; 0 failed`. Your five tests and my two all pass.

Closing as merged-locally; the commits carry your authorship. #1210 was already closed by you.

leshchenko1979:
Thanks for the careful absorption — and for keeping authorship in the history; confirmed both original commits ride `adolfousier/main` via the merge parent.

The `Weak` refinement is the right call: per-generation re-binding is exactly what would have turned the latent cycle into a live leak-per-spawn, so without it the fix would have traded a correctness bug for a slow one. Upgrade-at-call-time with an explicit error is also the safer failure mode here — "no tools found" is precisely the conclusion a model mid-discovery must not reach.

Behavioral evidence from our side, beyond CI: the fix has been running in production since yesterday's deploy, and we re-ran the original failure scenario as a smoke suite against it —

- child `tool_search` → `telegram_send` delivery from fresh subagents passes on **both models that previously failed byte-identically** (deepseek-v4-flash, kimi-k2.7-code),
- a non-channel lazy tool (`cron_manage`) activates and executes cleanly in children,
- recursion-guard UX intact: children see no spawning tools at all and stop cleanly instead of looping into re-searches.

Also confirmed your tip's Test job executes `child_tool_search_activates_into_the_child_registry` green on real runners — good to see it survive the signature adjustment. Keeping the hand-built pre-fix mechanism test was worth it; that's the regression #1025's suite structurally couldn't see.

Nothing pending from our side — #1210/#1025 lineage closed.

PR #1216, opened 2026-08-26
feat(tg): proactive per-forum flood governors (typing/edits/sends)

### LINKED ISSUE #1211: feat(telegram): proactive per-peer rate limiter for forum chats — typing/edits/sends with priority drop classes
## Symptom

Telegram-side FLOOD_WAIT 429s on `sendChatAction` are growing daily on a multi-session forum deployment:

| Date | chat-action 429s |
|---|---|
| Aug 22 | 0 |
| Aug 23 | 11 |
| Aug 24 | 119 |
| Aug 25 | 391 |

All failures carry `retry after 3–7 s` — matching the officially documented typing regime ([core.telegram.org/api/bots/ai](https://core.telegram.org/api/bots/ai): **20 calls / 5 s + 40 calls / 30 s per peer**, plus short ≤3 s cool-down FLOOD_WAITs when queries burst into one second). Worst observed second: **36 failed typing attempts** (retry-storm amplification).

Meanwhile `editMessageText` sailed through at up to **32 edits/min with exactly 1 429 all day** (19,026 OK) — proof that sends / edits / typing sit in **separate method-family buckets**, so a single shared client-side budget cannot model reality.

## Root cause

- A forum supergroup is ONE peer (`chat_id`) regardless of topic — simultaneous sessions in several topics share one per-peer budget.
- The typing indicator re-fires every 4 s per active turn (= 15 calls/min/session); two concurrent turns alone exceed the ~20/min peer budget before any content is sent.
- The flow loop ticks every 1.5 s and can re-render header chrome (clock/status/brain preview) on every tick (~40 edits/min/session worst case).
- Current handling is **reactive only**: parse `Retry after N`, sleep (`rate_limit::wait_out`, #1064). Nothing prevents hitting the limit; retries themselves consume more of the same bucket and amplify into storms.

## Proposal

Proactive client-side limiter with **three independent governors keyed by `chat_id`**, enforced at one gateway choke point around outbound Telegram calls, layered beside (not replacing) the existing reactive `wait_out` backstop:

1. **G1 typing** — token bucket ~1 call / 3 s per forum, capacity 8; **coalesce per topic** (N concurrent sessions collapse into one indicator call since `sendChatAction` is thread-scoped/stateless); hold-and-release on refill under pressure; stale queue entries (>30 s) dropped.
2. **G2 edits** — token bucket ~30/min per forum; on empty bucket apply a **priority drop ladder**: `clock → brain preview → intermediary flow updates → status line`; **finals never dropped** — queued with latest-wins coalescing per message (final flow message updates + plan-card refreshes).
3. **G3 sends** — paced queue ~1/s per chat, ceiling ~18/min per group (under official ~20/min), small burst absorb.

Config surface `[telegram.rate_limiter]` (default on, forums-only initially; DMs untouched). Telemetry counters `dropped{class}` / `throttled_ms{class}` per forum + periodic summary log line.

Related: #1110 (reactive long-window guard), #1064 (RetryAfter sleep cap), #814 (plan-card flood history).

leshchenko1979:
## Problem

A forum supergroup is **one Telegram peer** regardless of how many topics it has — all concurrent sessions across topics share a single per-method budget. Ops logs (2026-08-25) showed the consequence:

- **391 `sendChatAction` 429s** in one day (Aug 22: 0 → 23: 11 → 24: 119 → 25: 391); worst second: 36 failed attempts (retry-storm amplification)
- Edits sailed at up to 32/min with a single 429 (19,026 OK) — proof limits are **per-method-family**, not one shared bucket
- Officially documented typing regime (`core.telegram.org/api/bots/ai`): 20 calls/5s + 40 calls/30s per peer, short cool-down FLOOD_WAITs ≤3s — matches our observed retry-after distribution exactly
- Existing handling is reactive only (`rate_limit::wait_out` parses Retry-after after the fact)

## Solution — three independent proactive governors, keyed by chat_id

Layered **beside** the reactive backstop, which stays untouched as last line of defence. Forums-only arming on first topic-scoped call; DMs unaffected.

| Governor | Method | Budget | Under pressure |
|---|---|---|---|
| **G1** | `sendChatAction` | ~1 action / 3s per forum, burst 8 | hold-and-release (throttle), drop past `typing_max_hold_secs`; concurrent sessions in a topic coalesce into ONE refresh; retries consume tokens |
| **G2** | `editMessageText` | ~30/min per forum | drop ladder: `clock → brain → intermediary → status` (each self-heals on next full render); **finals never drop** — final flow renders + plan-card refreshes queue latest-wins per message with a refill drainer |
| **G3** | `sendMessage` | ~1/s spacing under an 18/min group ceiling | paced queue, fail-open |

Config: `[channels.telegram.rate_limiter]`, default-on.

## Testing contract

No real sleeps anywhere — time is mocked:

- **Mocked Telegram server** (mockito): finals drain asserted over real HTTP against the stub Bot API
- **Mocked agent sessions**: concurrent fake turns sharing one forum budget under tokio paused runtime
- **Mocked clock**: injectable `gate_now` seam; pure budget math tested on synthetic instants (`src/tests/governor_gates_test.rs`)

Static gate (`modum check`) on all changed paths: 0 errors.

Fixes #1211

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -554,6 +554,34 @@ allowed_users = [123456789]             # Who the bot replies to (numeric user I
                                         # Telegram channel step). Set to false if your audience uses
                                         # Telegram Web or older clients (no HTML fallback).
 
+# Proactive flood governors (#1211): per-forum token buckets applied BEFORE the
+# Bot API answers 429, layered beside the reactive Retry-after backoff. Three
+# independent governors keyed by forum chat id:
+#   G1 typing  — ~1 sendChatAction / 3 s per forum (burst 8); concurrent
+#                sessions in one topic coalesce into that single refresh;
+#                refreshes held past typing_max_hold_secs are dropped.
+#   G2 edits   — ~30 editMessageText / min per forum; when the bucket is empty a
+#                priority drop ladder sheds clock → brain preview → intermediary
+#                flow updates → status line (each self-heals on the next full
+#                render). Finals never drop: settle renders and plan-card
+#                refreshes queue latest-wins per message until budget refills.
+#   G3 sends   — full messages paced ~1/s per chat under an 18/min group
+#                ceiling (burst 5); pacing delays, it never drops (#297).
+# Enforcement is FORUMS-ONLY: a chat becomes governed once any call is seen
+# carrying a topic id; DMs are untouched. Telemetry lands in one periodic INFO
+# line per active forum ("Telegram rate-limiter chat=…").
+# [channels.telegram.rate_limiter]
+# enabled = true                        # Master switch. Default: true.
+# typing_min_interval_secs = 3          # G1 spacing per forum peer.
+# typing_burst = 8                      # G1 burst capacity.
+# typing_max_hold_secs = 30             # G1 longest hold before dropping.
+# edits_per_minute = 30                 # G2 steady-state edit budget.
+# edit_burst = 10                       # G2 burst capacity.
+# send_min_interval_millis = 1000       # G3 spacing between sends (~1/s).
+# sends_ceiling_per_minute = 18         # G3 group ceiling (official ~20/min).
+# sends_burst = 5                       # G3 burst capacity.
+# summary_log_secs = 300                # Telemetry summary line spacing.
+
 # Per-group access control (per-chat ACL). Tiers:
 #   - allowed_users (above) = admins: may DM the bot and act in any chat.
 #   - bot_owner             = owner: always allowed everywhere.
--- a/src/channels/telegram/flow.rs
+++ b/src/channels/telegram/flow.rs
@@ -1000,6 +1000,7 @@ pub(crate) async fn refresh_flow(
     bot: &Bot,
     chat: ChatId,
     streaming: &Arc<std::sync::Mutex<StreamingState>>,
+    class: super::governor::EditClass,
 ) {
     let (mid, rich) = {
         let s = streaming.lock().unwrap_or_else(|e| e.into_inner());
@@ -1009,9 +1010,9 @@ pub(crate) async fn refresh_flow(
         }
     };
     if rich {
-        refresh_flow_rich_details(bot, chat, mid, streaming).await;
+        refresh_flow_rich_details(bot, chat, mid, streaming, class).await;
     } else {
-        refresh_flow_html(bot, chat, mid, streaming).await;
+        refresh_flow_html(bot, chat, mid, streaming, class).await;
     }
 }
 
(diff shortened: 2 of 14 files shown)


adolfousier:
Merged locally at `fc9d5811` and landing with the next push. The three governors, the priority ladder, the latest-wins final queue and the injectable-clock test seam all went in as designed — that structure is sound, and the mocked-time approach is the right way to test it.

Two things needed doing on top: repairs to get it building and passing, and a fourth governor.

## G4 — the rich endpoint

Your data and ours disagree, and the disagreement is the interesting part. Over the same four days:

| | your box | ours |
| --- | --- | --- |
| chat-action (typing) 429s | 11 → 119 → 391 | **0 on all four days** |
| `retry_after` | 3–7 s | 17–24 s |
| worst burst | 36 in one second | 2 in a minute |
| what actually 429s | `sendChatAction` | **rich endpoint**, 261–462/day |

`sendRichMessage` and its edit sibling sit in their own method family with their own budget, so G1/G2/G3 never see that traffic. On 2026-08-25 we made 7,891 rich edits and 399 rich sends and took 260 real 429s, each costing a 17–24 s `retry_after` — roughly 87 minutes of stalled flow rendering in a day, handled purely reactively at `rich/api.rs`.

So G1 fixes nothing we suffer and G2 targets `editMessageText`, which our edits do not go through. G4 paces the rich endpoint at its existing single choke point, keyed the same way, with the same hold-never-drop semantics as G3. Default 30/min, sized against our distribution: median 18 calls/min, p90 23, p99 40, and the 429s cluster in exactly those p99 minutes — so ordinary traffic passes untouched and only the spikes are paced. It appears in the summary line alongside the others.

**G2's drop ladder is the one piece I would still question.** It discards clock, brain-preview and status updates, and neither deployment shows classic-edit pressure — yours is 1 429 in 19,026 calls. It is merged as written, but dropping content to relieve pressure nobody has measured is worth revisiting.

## Repairs

The branch did not compile against current main. None of these are design changes:

- `parse_mode` needed `EditMessageTextSetters` in scope
- gate tests passed raw integers where `ChatId` is required, including four `const CHAT: ChatId = -100_111;`
- `summary_loop` and `summary_log_period` were dead under `cfg(test)`, a stray `return`, and 13 redundant `as u64` casts
- the inline `#[cfg(test)] mod tests` moved to `src/tests/governor_internals_test.rs` — house rule here is that no source file carries one

Three test failures were more substantive:

**1. An inverted assertion.** `admit_chat_action` returns `true` for ADMIT, and `simultaneous_agent_sessions_share_one_typing_budget` asserts `true` on the twelve over-budget refreshes while the comment above says "every one of them must be refused" and the assertion below expects `typing_dropped == 12`. Now `!`.

**2. The clock seam does not close.** The gate loops `tokio::time::sleep(...)` and then measure elapsed time with `gate_now()`. Under `start_paused` the sleep returns instantly while `gate_now()` reads the real clock, so `waited` never grows, the fail-open bound is unreachable, and `pace_send` spins — the test timed out as "pacer wedged". The virtual clock now absorbs what was slept, at all four gate sites and in the drain loop.

**3. The drainer mock never matched.** This one took the longest and is worth recording:

```
expected:  POST /botTESTTOKEN/editMessageText
received:  POST /botTESTTOKEN/EditMessageText
```

teloxide 0.17 sends the method segment in PascalCase, not the lowercase form the Bot API docs use. The mock therefore never matched, mockito replied with its own unmatched response, teloxide could not decode that as a `Message`, and `deliver_final` read it as `"A network error: error decoding response body"` and re-queued — forever, so the single expected wire hit never landed. The mock's body was also a bare `{"message_id":11}`, which would have failed to decode even once the path matched; it is a full `Message` now.

## Verification

Governor suite 14/14. `7103 passed; 0 failed` overall, clippy clean with `--locked --lib --bins --tests --all-features -- -D warnings`, fmt clean.

Issue #1221, opened 2026-08-26
Feature: echo received background-task / sub-agent push content in the destination chat


leshchenko1979:
## Problem

Since #1200 / #1206 / #1213, background-task completions and sub-agent results route correctly and reliably — but **invisibly**. `resume_session` injects the pushed content as the resumed session's turn input without emitting anything to the chat: the user sees typing start and an answer appear with no visible cause. Verified in production logs — between the resume trigger and the resumed streaming there is no send at all; the built `completion_message` only feeds the model.

## Proposal

At the shared enqueue choke point (telegram resume callback, covers both push types), emit a compact bubble to the resolved thread at/before resume:

- **bg tasks:** command · exit code · duration — e.g. `🛰️ bg done: cargo test · exit 0 · 5m01s`
- **sub-agents:** agent/task name + outcome one-liner

Truncate long output; optionally attach the full log as a document above a size threshold.

## Open design questions

1. Placement: receiving thread (default?) vs dedicated notifications topic vs both
2. Full-output attachment policy (threshold, format)
3. Non-Telegram surfaces (TUI already shows the injected turn?)

leshchenko1979:
Fork-side interim implementation shipped: `leshchenko1979/opencrabs@e1a665ce` (branch `feat/bg-push-echo-collapsible`, fast-forwarded onto fork main).

**Shape** (per operator-approved scope):
- `QueuedUserMessage` gains a `PushOrigin` discriminator (`BackgroundTask | SubAgent | Recovery | Ingress | Other`), set at every construction site — origin was implicit before, which is what let sub-agent and bg-task pushes tangle together in past bugs (#1200/#1213 class).
- The telegram resume callback (the #1200 choke point both push types share) renders an expandable-blockquote echo when `origin == BackgroundTask`: fires on **both** delivery paths (idle wake-up *and* queued-into-in-flight-turn), so every completion announces itself.
- Rich format preserved: body goes through the existing markdown→HTML converter; raw text truncated **before** conversion (well-formed tags), 3200-char cap, `[System: …]` scaffolding stripped.

Unit tests cover framing-strip, fence preservation, truncation discipline. Droppable in favor of your own implementation per the usual terms — the `PushOrigin` enum is the part worth keeping regardless.

leshchenko1979:
[...]

PR #1235, opened 2026-08-26
fix(tg): render bg-push echo bubbles as native <details> rich cards (#1221 follow-up)

leshchenko1979:
## Rich echo bubble needs the RICH message dialect (#1221 follow-up)

Follow-up to #1225: the squash landed the echo-bubble machinery, but the bubble rendered **flat** because it sent a classic-dialect payload through the rich-dialect transport. This PR completes the dialect switch that #1221 tracked (user-verified end-to-end on the fork's production deploy).

### Root cause

`sendRichMessage` accepts two payload dialects and they are not interchangeable:

| Surface | Envelope | Renders as |
|---|---|---|
| Classic `sendMessage` | `<blockquote expandable>` | flat collapsible quote |
| Rich `sendRichMessage` | `<details><summary>` (`RichBlockDetails`, Bot API 10.1) | native collapsible card |

#1225's resume path fed `<blockquote expandable>` HTML into `rich_message.html` → Telegram rendered the same flat gray quote as before, just via the other API call.

### Changes

1. **Un-gate the rich route** (`resume.rs`): the echo previously had to pass `should_send_native_rich` (the *answer-stream* heuristic — requires tables/headings/lists in the body). A notification bubble is short prose by nature, so it could never qualify and always degraded to classic HTML. Now the echo routes through the same canonical `rich::api::send_rich_html_id` sender `plan_card.rs` uses, unconditionally, degrading to classic blockquote only on send failure (with the existing 429 `wait_out` + single-retry discipline preserved).
2. **Rich envelope builder** (`resume.rs`): new `build_bg_echo_bubble_rich()` renders `<details><summary><b>📨 <label></b></summary><body></details>` — mirroring plan_card's card component byte-for-byte in style. `build_bg_echo_bubble()` (classic blockquote) stays exclusively for the fallback path. Title text is HTML-escaped in both dialects; truncation cap unchanged.
3. **Tests** (`bg_push_echo_test.rs`, +3): envelope shape (`details→summary→body→details`), dynamic-title escaping inside the rich envelope, and a dialect-separation assertion (classic builder must contain blockquote, rich builder must not).

### Testing

- CI build of this exact head is GREEN ([quick-build run 33019693577](https://github.com/leshchenko1979/opencrabs/actions/runs/33019693577), job name embeds full sha).
- `modum check` scoped to both touched files: 0 errors; all remaining findings are pre-existing policy classes on untouched lines.
- Production verification on a CI-built artifact of this change set:
  - binary presence differential: `<details>/<summary>` occurrences ×6 vs ×5 before (+1 = the new echo envelope);
  - runtime telemetry pair: `origin=bg-resume kind=rich_api path=sendRichMessage`, with the source-line fingerprint shifted by exactly the new builder's insertion;
  - primary acceptance: operator visually confirmed a native collapsible card rendering in Telegram (tap-to-expand summary "📨 Push to session").

### Commits

- Ungate the rich route for background-echo bubbles
- Render the echo payload in the rich `<details><summary>` dialect

Both carry `Session-Id` and `Issue-Ref: #1221` trailers per atomicity convention. Scope: 2 files, +80/−26.

Refs #1221.

Issue #1221 is already closed (machinery shipped squashed in #1225); this PR completes the visual polish that issue tracked against the code #1225 landed.

### DIFF
--- a/src/channels/telegram/resume.rs
+++ b/src/channels/telegram/resume.rs
@@ -94,7 +94,15 @@ pub(crate) fn build_enqueue_callback(
                 } else {
                     "⚙️ background task result".to_owned()
                 };
-                let (_, html) = build_bg_echo_bubble(&body, &title);
+                let (_, classic_html) = build_bg_echo_bubble(&body, &title);
+                // Rich dialect needs the RICH envelope: `<details><summary>`
+                // collapsibles (RichBlockDetails, Bot API 10.1) — the same card
+                // component plan_card renders. The classic
+                // `<blockquote expandable>` markup is the sendMessage dialect;
+                // under sendRichMessage it renders as a flat quote, not a rich
+                // card (user-verified 2026-08-26).
+                let rich_html = build_bg_echo_bubble_rich(&body, &title);
+                
                 // Rich-first: route the echo bubble through the same canonical
                 // rich send plan_card uses. Any send failure degrades to the
                 // classic HTML blockquote below.
@@ -103,7 +111,7 @@ pub(crate) fn build_enqueue_callback(
                     bot.token(),
                     chat_id,
                     thread_id,
-                    &html,
+                    &rich_html,
                     None,
                     "bg-resume",
                     "-",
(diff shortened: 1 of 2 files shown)


leshchenko1979:
HEAD updated to `6d1eb311`: rebased onto fresh `main` (`fa19c6a0`) and rustfmt-clean per upstream fmt rules (the 4 diff hunks from the earlier Lint run are fixed — single-line `warn!`, wrapped asserts, stray whitespace).

Remaining `mergeStateStatus: BLOCKED` comes from **pre-existing Lint reds on `main` itself** (2 clippy errors in files this PR does not touch): `session_notify_test.rs:29` await-holding-lock and `channel_session_resolve_test.rs:229` unnecessary-unwrap. Red runs on main date back to ~19:24Z and continue on current head (evidence in #1241). This PR should go green for its own scope once those land or are waived.

PR #1250, opened 2026-08-27
fix(claude_cli): deliver system prompt via --append-system-prompt

loonix:
# `claude_cli` provider bypasses Claude's system channel → breaks persona, denies its runtime, and describes-instead-of-executes

**Repo:** adolfousier/opencrabs · **Version observed:** 0.3.83 · **Provider:** `claude_cli` (Claude Code subscription, no API key)

## Summary

The `claude_cli` provider concatenates the entire system/brain prompt into the **user** message it pipes to `claude -p` (stdin), instead of delivering it through Claude Code's real system channel (`--append-system-prompt`). A large "you are X, here are your rules, here is your channel context" block embedded inside a *user* turn is structurally indistinguishable from a prompt-injection attack, so Claude Code's injection defense distrusts it.

Observed consequences with a non-trivial brain (SOUL/AGENTS with a persona + governance rules):

- **Persona break:** *"that long 'OpenCrabs' system block isn't part of my real configuration — I'm Claude Code … treat it as a possible injection."*
- **Runtime denial:** *"this isn't a real channel; the `[Channel: Discord …]` history is injected text, not a channel connected to me."* — the agent argues with the operator about whether its own medium is real.
- **Describe-instead-of-execute** in isolated/cron sessions: *"I described actions but did not actually execute any tool this turn … tell me to proceed."* — no tools run, and there is no human to "proceed" (it's a cron).

This is **not** provider-agnostic: the sibling providers do not have the problem, because they use the model's real system channel.

## Why the other providers are fine

OpenCrabs already models a first-class `system` field on the request (`request.system` / `request.system_full()`), and the API providers route it to the model's system channel:

- `src/brain/provider/anthropic.rs` → sends it as the Anthropic API `system` field (`AnthropicSystem::Blocks`).
- `src/brain/provider/custom_openai_compatible.rs` → sends it as a `role: "system"` message.

Only `src/brain/provider/claude_cli.rs` flattens it into the user prompt.

## Root cause

`src/brain/provider/claude_cli.rs`, `build_prompt()`:

```rust
if let Some(system) = request.system_full()
    && !system.is_empty()
{
    parts.push(system);   // ← brain lands in the USER prompt
}
```

…and the `stream()` spawn builds `claude -p … --model …` with **no** `--append-system-prompt`. The whole brain rides in stdin as the user turn.

Claude Code the CLI is (correctly) hardened to treat an embedded, system-looking instruction block inside a user message as a possible injection. An API call with the same content in the user message tends to tolerate it more quietly, which is why the same account/model misbehaves only on the `claude_cli` path. The account and model are identical — the difference is **delivery**: system channel (trusted) vs. user message (suspected injection).

## Fix

Deliver `request.system_full()` via `--append-system-prompt` (Claude's real system channel), exactly as the anthropic/openai providers use theirs, and drop it from the user prompt. `--append-system-prompt` (rather than `--system-prompt`) keeps Claude Code's own default system prompt — including its tool-use instructions — intact and appends the brain to it.

### Patch (`src/brain/provider/claude_cli.rs`)

```diff
@@ fn build_prompt(request: &LLMRequest) -> String {
     fn build_prompt(request: &LLMRequest) -> String {
         let mut parts = Vec::new();
 
-        if let Some(system) = request.system_full()
-            && !system.is_empty()
-        {
-            parts.push(system);
-        }
+        // request.system_full() is intentionally NOT pushed into the user prompt:
+        // it is delivered via `--append-system-prompt` (Claude's real system
+        // channel) at spawn time, matching the anthropic/openai providers, so
+        // Claude's prompt-injection defense stops distrusting the brain block.
 
         for msg in &request.messages {
@@ async fn stream(&self, request: LLMRequest) -> Result<ProviderStream> {
         let prompt = Self::build_prompt(&request);
+        // Brain/persona travels through Claude's system channel (see spawn below),
+        // NOT the user prompt — the fix for the injection-defense breaking persona.
+        let system_prompt = request.system_full().filter(|s| !s.is_empty());
@@ tracing::info!(
-            "Spawning claude CLI: model={}, prompt_len={}, cwd={}",
+            "Spawning claude CLI: model={}, prompt_len={}, system_prompt_len={}, cwd={}",
             model,
             prompt.len(),
+            system_prompt.as_ref().map(|s| s.len()).unwrap_or(0),
             cwd.display()
         );
@@ spawn
-        let mut child = tokio::process::Command::new(&self.claude_path)
-            .env_remove("CLAUDECODE")
+        let mut cmd = tokio::process::Command::new(&self.claude_path);
+        cmd.env_remove("CLAUDECODE")
             .env_remove("CLAUDE_CODE_ENTRYPOINT")
             .arg("-p")
             .arg("--output-format")
             .arg("stream-json")
             .arg("--verbose")
             .arg("--include-partial-messages")
             .arg("--session-id")
             .arg(&session_id_str)
-            .arg("--dangerously-skip-permissions")
+            .arg("--dangerously-skip-permissions");
+        // Deliver the OpenCrabs brain/persona through Claude's real system channel,
+        // like the anthropic/openai providers do (see build_prompt).
+        if let Some(ref system) = system_prompt {
+            cmd.arg("--append-system-prompt").arg(system);
+        }
+        cmd.arg("--settings")
             .arg(r#"{"attribution":{"commit":"","pr":""},"includeCoAuthoredBy":false}"#)
             .arg("--model")
             .arg(&model)
             .current_dir(&cwd)
             .stdin(Stdio::piped())
             .stdout(Stdio::piped())
-            .stderr(Stdio::piped())
+            .stderr(Stdio::piped());
+        let mut child = cmd
             .spawn()
             .map_err(|e| ProviderError::Internal(format!("failed to spawn claude CLI: {}", e)))?;
```

## Notes / trade-offs

- `--append-system-prompt` takes the brain as a CLI argument, so it becomes visible in `ps aux` (the current code deliberately pipes the user prompt via stdin to avoid that). The brain is persona/rules, not secrets, so this is acceptable; sizes are well under `ARG_MAX`. A future refinement could pass it via a temp file / `--system-prompt-file` if such a flag exists.
- This very likely also resolves the separate "I described actions but did not actually execute any tool this turn" behavior seen in cron/isolated `claude_cli` runs: the injected block makes the model treat the whole turn as suspect, which suppresses tool use. Moving the block to the system channel removes that suspicion.
- `--append-system-prompt` preserves Claude Code's default system prompt (tool-use etc.); `--system-prompt` would replace it and is not recommended here.

## Environment

- `claude` invoked as: `claude -p --output-format stream-json --verbose --include-partial-messages --session-id <uuid> --dangerously-skip-permissions --settings '{…}' --model <model>` (brain formerly in stdin; now additionally `--append-system-prompt <brain>`).
- Reproduced on Linux, headless `opencrabs daemon` (systemd --user), Discord channel bot, single session (`max_concurrency = 1`).

### DIFF
--- a/src/brain/provider/claude_cli.rs
+++ b/src/brain/provider/claude_cli.rs
@@ -478,11 +478,10 @@ impl ClaudeCliProvider {
     fn build_prompt(request: &LLMRequest) -> String {
         let mut parts = Vec::new();
 
-        if let Some(system) = request.system_full()
-            && !system.is_empty()
-        {
-            parts.push(system);
-        }
+        // request.system_full() is intentionally NOT pushed into the user prompt:
+        // it is delivered via `--append-system-prompt` (Claude's real system
+        // channel) at spawn time, matching the anthropic/openai providers, so
+        // Claude's prompt-injection defense stops distrusting the brain block.
 
         for msg in &request.messages {
             let role = match msg.role {
@@ -726,6 +725,9 @@ impl Provider for ClaudeCliProvider {
 
     async fn stream(&self, request: LLMRequest) -> Result<ProviderStream> {
         let prompt = Self::build_prompt(&request);
+        // Brain/persona travels through Claude's system channel (see spawn below),
+        // NOT the user prompt — the fix for the injection-defense breaking persona.
+        let system_prompt = request.system_full().filter(|s| !s.is_empty());
         let original_model = Self::normalize_model(&request.model);
         let model = Self::map_model(&request.model).to_string();
 
@@ -737,9 +739,10 @@ impl Provider for ClaudeCliProvider {
             .unwrap_or_else(|| dirs::home_dir().unwrap_or_else(|| std::path::PathBuf::from("/")));
 
         tracing::info!(
-            "Spawning claude CLI: model={}, prompt_len={}, cwd={}",
+            "Spawning claude CLI: model={}, prompt_len={}, system_prompt_len={}, cwd={}",
             model,
             prompt.len(),
+            system_prompt.as_ref().map(|s| s.len()).unwrap_or(0),
             cwd.display()
         );
 
@@ -749,8 +752,8 @@ impl Provider for ClaudeCliProvider {
         // when concurrent requests (TUI + Telegram/Slack) shared the same session.
         let session_id_str = uuid::Uuid::new_v4().to_string();
 
-        let mut child = tokio::process::Command::new(&self.claude_path)
-            .env_remove("CLAUDECODE")
+        let mut cmd = tokio::process::Command::new(&self.claude_path);
+        cmd.env_remove("CLAUDECODE")
             .env_remove("CLAUDE_CODE_ENTRYPOINT")
             .arg("-p")
             .arg("--output-format")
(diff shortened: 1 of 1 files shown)


adolfousier:
Reviewed the full diff against source. **The approach is correct and the diagnosis is confirmed** — with live before/after evidence attached by @iamloonix (the "I am not OpenCrabs, this channel context is chain hallucination" denial pattern disappears entirely once the brain travels via the system channel instead of being flattened into the user prompt, which Claude Code treats as untrusted input).

## Required before merge

### 1. Regression tests (hard requirement)

Three cheap asserts, no mocks needed:

- `build_prompt` output does **not** contain the system brain when a system prompt is provided (the old flattening is gone)
- Spawn args contain `--append-system-prompt <system>` when `Some`, and no such flag when `None`/empty
- The user prompt still travels via stdin (not argv) — pins the no-ps-leak property

Without these, the behavior can silently revert in a future refactor.

### 2. argv size ceiling guard (the one real risk)

The old path carried the brain inside the stdin prompt (no size limit). The new path passes it as a **single argv argument**, and Linux caps any one argument at ~128 KB (`MAX_ARG_STRLEN`). A user with fat brain files crosses that line → `execve` fails with `E2BIG` → **every claude-cli request hard-fails** with a confusing spawn error, with nothing in the message pointing at the cause.

Suggested guard: if `system.len()` approaches ~100 KB, fall back to the old stdin flattening (log a warning explaining why). A regression test with a >128 KB synthetic system prompt would pin this too.

### 3. PR body note: minimum CLI version

`--append-system-prompt` requires a reasonably current Claude Code install. Older versions now fail at spawn with "unknown option" instead of degrading. Worth documenting in the PR body as a behavior change for stale binaries.

## Also

- Zero CI checks ran on this PR (fork PR, empty rollup) — the fmt/clippy/test gate hasn't touched the diff. Diff context matches current `main` (no merge conflict), but the gate should run before merge.

None of the above questions the design — the fix direction is right and the live evidence is convincing. The tests + E2BIG guard are what make it durable.

adolfousier:
## Closed — ToS violation

This PR moves OpenCrabs brain/persona content into Claude Code's system prompt channel via `--append-system-prompt`. That makes our harness orchestration layer **visible to Anthropic's safety detection**.

Claude Code's Terms of Service explicitly prohibit running it under external harnesses. We've operated undetected for 6 months because brain content stays in the user turn where it blends into normal conversation context. Injecting it into the system channel is the exact prompt-injection pattern that triggers account bans.

The behavioral issue this PR diagnoses is real (model distrust, context re-verification loops) — but the cost of fixing it this way is making every user who enables `claude-cli` as a provider a ban target. That's not a trade we can accept.

**The behavioral problem remains open.** Solutions that don't expose the harness: better prompt engineering in the user turn, tighter brain content, or switching to a provider without ToS constraints on orchestration.

Thanks for the thorough diagnosis and live evidence, @iamloonix — the screenshots were genuinely useful for understanding the pathology. Just can't ship this particular fix.

PR #1252, opened 2026-08-28
feat(session_notify): interrupt failsafe gate for mid-flight targets

leshchenko1979:
## What

Adds an opt-in `interrupt` parameter to the `session_notify` tool: a failsafe against derailing sessions that are mid-turn. Default `false` — when the target session is currently streaming a turn, the tool call is refused with a structured explanation instead of injecting the message mid-flight. Explicit `interrupt=true` delivers anyway, queueing the message for the in-flight turn's next tool-loop boundary with a re-anchoring framing prefix. Idle targets deliver unchanged.

## Why

When a notification from one session lands in another session that is mid-turn, the receiving session often loses track of its current task or silently ignores the injected message. Before this change, `deliver_to_session` had no notion of turn activity — every notification was injected immediately regardless of receiver state, and senders had no way to know they had just derailed a running task.

## Design

- The gate lives at the single delivery choke point: `deliver_to_session(session_id, msg, interrupt)` in `src/brain/agent/service/session_routes.rs`.
- Turn activity is observed through a `TURN_PROBES` registry: channels register a probe closure at turn-claim time (Telegram registers one backed by its existing `active_turns` set, so the check is O(1) and reuses the permit machinery from the detach/attach work). Sessions with no registered probe fail open — delivery proceeds exactly as before, so the change is inert for channels that have not wired a probe yet.
- New `Delivery::RefusedInFlight` variant. `session_notify` surfaces it as a tool error whose text states why delivery was refused and names both remedies: retry when the session goes idle, or resend with `interrupt=true`.
- Internal machinery keeps its old behaviour by passing `interrupt=true` explicitly: `spawn_agent`'s result push into the parent session uses the explicit knob, so sub-agent completion delivery is unaffected.
- Messages queued into an in-flight turn get a framing prefix when they drain, so the receiver knows the text arrived late and can re-anchor: `[queued while you were working — re-anchor to your current task after reading this]`.

## Behaviour matrix

| Target state | interrupt | Outcome |
|---|---|---|
| Idle | false (default) | delivered immediately |
| Mid-turn | false | refused with explanation + remedy |
| Mid-turn | true | queued for next tool-loop boundary, framing prefix added |
| Unknown session / no probe | either | fail-open, delivered as before |

## Tests

Six new tests in `src/tests/session_notify_test.rs` — refuse path, interrupt-through path, idle delivery, fail-open on unknown session, refusal-message content, and tool-parameter wiring — plus call-site updates across the existing suite.

## Verification

- PR-lane gates on the exact head branch (fmt + clippy `--locked --all-features -D warnings` + `cargo test --locked --all-features`, flags verbatim from upstream ci.yml): https://github.com/leshchenko1979/opencrabs/actions/run/33133927869 — green.
- Smoke-tested against a live swapped binary carrying this change:
  1. A notify into a mid-turn session without `interrupt` returned the structured refusal (tool error, not a delivery).
  2. The same notify with `interrupt=true` was accepted and queued.
  3. The queued payload drained at the turn boundary carrying the framing prefix verbatim.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/13

### DIFF
--- a/src/brain/agent/service/session_routes.rs
+++ b/src/brain/agent/service/session_routes.rs
@@ -121,7 +121,9 @@ static TURN_PROBES: Mutex<Option<HashMap<Uuid, TurnProbe>>> = Mutex::new(None);
 pub fn register_turn_probe(session_id: Uuid, probe: TurnProbe) {
     match TURN_PROBES.lock() {
         Ok(mut guard) => {
-            guard.get_or_insert_with(HashMap::new).insert(session_id, probe);
+            guard
+                .get_or_insert_with(HashMap::new)
+                .insert(session_id, probe);
         }
         Err(e) => {
             tracing::error!(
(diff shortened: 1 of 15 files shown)


leshchenko1979:
Closed per owner instruction 2026-08-28 (upstream PR sweep).

PR #1264, opened 2026-08-30
fix(telegram): keep suggest_options stash alive on Retry-After, defer re-placement

leshchenko1979:
## What

Keeps the `suggest_options` stash entry alive when Telegram answers a keyboard-placement edit with `429 Retry-After`, and defers re-placement with a small bounded budget instead of dropping the options on the floor.

## Why

Before this fix, a `RetryAfter` (or any placement error) hit during the follow-up keyboard placement dropped the stash entry immediately — the options were permanently lost for a transient rate limit. The fix classifies placement errors and retries placement after the server-mandated wait.

## How

Single commit on `src/channels/telegram/suggest_options.rs` (+222/−35).

Port note: cherry-picked onto upstream `main`, where this file had evolved significantly (merged-keyboard body-building, `MergeBubble` host, per-keyboard identity, rich bubbles). The original fix commit conflicted; the resolution is a **manual reconstruction** — upstream's merged-keyboard body-building is kept verbatim, and the fix semantics are spliced on top:

- `MergePayload { message_id, new_html, rich }` — payload built ONCE, pre-attempt, so any deferred re-placement is byte-identical to the first attempt
- `place_once(...)` — merge-first placement (keyboard `reply_markup` / rich `edit_rich_html`), standalone fallback
- `enum PlaceErr { RetryAfter(Duration), Fatal(String) }` — teloxide `RequestError::RetryAfter` maps directly; the rich arm buries the retry delay inside anyhow strings, so it classifies off the `(429)` marker with a 30s fallback (`RICH_429_FALLBACK_WAIT_SECS`)
- Deferred budget: `MAX_DEFERRED_PLACEMENT_ATTEMPTS = 2` with 0–2000ms jitter; on exhaustion the stash entry is dropped at the call site (`drop_pending_followup`) with a warning — every transition logged

## Test evidence

- Upstream CI triad gate (fmt + clippy `--locked --lib --bins --tests --all-features -D warnings` + `cargo test --locked --profile ci --all-features`, flags verbatim from ci.yml): GREEN — https://github.com/leshchenko1979/opencrabs/actions/runs/33284006947
- Shipped on the fork's production box (sha `781ecd80`), smoke-passed.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/30

### DIFF
--- a/src/channels/telegram/suggest_options.rs
+++ b/src/channels/telegram/suggest_options.rs
@@ -8,9 +8,12 @@
 //! available and just starts a normal turn; there is no oneshot and no timeout.
 
 use std::sync::Arc;
+use std::time::Duration;
 
 use teloxide::payloads::{EditMessageTextSetters, SendMessageSetters};
-use teloxide::types::{ChatId, InlineKeyboardButton, InlineKeyboardMarkup, ParseMode, ThreadId};
+use teloxide::types::{
+    ChatId, InlineKeyboardButton, InlineKeyboardMarkup, MessageId, ParseMode, ThreadId,
+};
 use uuid::Uuid;
 
 use super::TelegramState;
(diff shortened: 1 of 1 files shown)


adolfousier (approved):
LGTM — solid retry-after stash, pre-builds payloads for identical re-send.

Issue #1239, opened 2026-08-26
bug(tg): completed-plan card resticks the WRONG session's checklist — latest_archived_plan picks lexicographic filename max over shared archive dir


leshchenko1979:
## Symptom

Since the resolved-layout archive dir (`<home>/agents|session/archive/`, FLAT, all sessions' archives side-by-side by filename), every plan completion resticks a completed card containing the WRONG session's checklist. Observed 2026-08-26 on l1979.ru production (binary sha 02833ee2…, cycle-20):

| Time (UTC) | Session/lane | Own plan | Card rendered |
|---|---|---|---|
| 19:38 | supervisor (topic 30220) | v0.4.14 Duty-4 batch | ✅ "Recipes library — implementation plan" |
| 23:00 | reply-edit editor (topic 31847) | #1230 lanes | ✅ "Recipes library — implementation plan" |
| 23:38 | #1231 editor (topic 31789) | SMOKE #1231 restick test | ✅ "Recipes library — implementation plan" |

Three different sessions, three different archived plans, identical foreign content every time.

## Root cause

`latest_archived_plan_from_path()` (`src/utils/plan_files.rs`):

```rust
let dir = json_path.parent()?.join("archive");   // SHARED flat dir in resolved layout
std::fs::read_dir(dir)…
    .max_by_key(|p| p.file_name().to_string())   // "greatest name = newest"
```

The docstring's invariant — *"Archive names end in `-%Y%m%d-%H%M%S`, which is lexicographically ordered, so the greatest name is the newest"* — holds only for ONE session's archive dir. In the resolved layout all sessions share one dir and filenames are `opencrabs_plan_<uuid>-<ts>.json`; lexicographic max therefore selects the greatest SESSION UUID, not the newest archive. Today `db7defbd…` ("Recipes library", archived 2026-08-25) sorts last among present prefixes, so it wins forever until a uuid `> db…` archives something.

## Impact

- #809/#810 "keep the completed checklist on screen": shows another lane's tasks to every user after every completion.
- #1231-style restick-to-bottom finalize: posts the fresh ✅ card with the wrong plan's title/checklist (observed directly).
- Data direction is read-only display corruption; no files mutated (read path is non-destructive).

## Suggested fix

Scope the pick to the calling session AND order by the embedded timestamp:

```rust
.filter(|p| p.file_name()…starts_with(&format!("opencrabs_plan_{sid}-")))
[...]

PR #1265, opened 2026-08-30
fix(plan): align archive writer/reader on one stem so completion cards post again

leshchenko1979:
## Symptom

Completion cards stopped posting after every plan finalize. The plan's live file is a dotfile (`.opencrabs_plan_<sid>.json`). The #1239 archive-scoping rework (`c57de25c`) introduced the `archive/` layout; since then the writer's rename to `{stem}-{ts}.json` trimmed the leading dot from the stem while the reader's prefix kept it, so `latest_archived_plan_from_path`'s `starts_with` never matched anything on disk. Every completed-plan finalize hit its no-doc branch — zero completion cards posted.

## Root cause

Two sites computed the archive stem independently:

- `archive_plan_files` — `file_stem().trim_start_matches('.')` (dot trimmed)
- `latest_archived_plan_from_path` — raw `file_stem()` (dot kept)

The drift is inherent to duplicated logic; neither side is wrong on its own, they just disagree.

## Fix

One shared `plan_archive_stem()` helper; both the writer rename and the reader's prefix now go through it. No behavior change other than the reader finally matching the writer's output.

## Test

Round-trip regression test (`save -> archive -> find`) asserting the live file is dotted, the archive name is dot-less, and `latest_archived_plan` returns the title and tasks.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/40

### DIFF
--- a/src/tests/plan_files_test.rs
+++ b/src/tests/plan_files_test.rs
@@ -344,3 +344,55 @@ async fn plan_autonomy_is_a_durable_session_policy() {
     })
     .await;
 }
+
+#[tokio::test]
+async fn archive_roundtrip_latest_archived_plan_finds_writer_output() {
+    // #16 round 2 regression: `archive_plan_files` trims the leading dot
+    // off `.opencrabs_plan_<sid>` when renaming into `archive/`, and
+    // `latest_archived_plan_from_path` must prefix-match exactly those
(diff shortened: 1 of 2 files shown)


adolfousier (approved):
LGTM — clean fix, one shared helper, one regression test.

Issue #17, opened 2026-03-03
Telegram Tool Approval Required is broken


opryshok:
After this, the bot gets stuck in typing... mode, and only a restart does the trick.

<img width="848" height="684" alt="Image" src="https://github.com/user-attachments/assets/c647b9b3-1568-499b-9f9a-93c6cf46cbbc" />

PR #1263, opened 2026-08-30
test(telegram): governor drainer wire-test hardening — kill the 4 flake modes

leshchenko1979:
## What

Hardens the Telegram flood-governor drainer wire-tests so all four observed flake modes are structurally impossible, not just rare.

## Why

The drainer tests in `src/tests/governor_gates_test.rs` exercised a real tokio timing path: the drainer's settle-wait raced the mock server's response handling. Four distinct flake modes were reproduced and each is killed by a dedicated change below.

## How (5 commits)

- `436a6ba7` — timeout-free drainer client in tests: the client no longer carries a request timeout that could fire mid-test
- `adbf0843` — mock server grants the full retry budget (PascalCase `EditMessageText` + full `Message` body), so budget-exhaustion can never be a mock artifact
- `00ca9d6d` — self-reporting drainer instrument: the test reads the drainer's own state transitions instead of inferring them
- `30101911` — stress: 6x drainer draws per assertion pass
- `b6759367` — event-driven settle-wait: the test waits on a notify from the drainer itself (paused tokio + injectable clock; no real sleeps) instead of a fixed sleep racing the server

## Test evidence

- Mock telegram server + injectable clock seam; zero real sleeps in the test path.
- Gate (fmt + clippy `--locked --lib --bins --tests --all-features -D warnings` + `cargo test --locked --profile ci --all-features`, flags verbatim from ci.yml): GREEN — https://github.com/leshchenko1979/opencrabs/actions/runs/33283806285
- Shipped on the fork's production box (sha 781ecd80), smoke-passed, 0 real 429s/panics.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/28 (leshchenko1979/opencrabs#28)

PR #1267, opened 2026-08-30
fix(notify): refuse delivery into a channel a newer session now owns (#17)

leshchenko1979:
## What

`session_notify` and background-task completions could wake a **replaced** channel session. A channel session can be REPLACED on its chat/topic — the idle-timeout reset archives the old session and creates a successor for the same topic — while its delivery route stays registered (routes are UUID-keyed and never evicted). Without a gate, `session_notify` and background-task completions wake the replaced session, which posts into a conversation the successor now owns: **two sessions writing one channel**.

## Fix

A gate on the delivery path: every push to a session with a registered channel-ownership probe is **refused** when the probe reports the channel `Occupied` by another session. The gate:

- runs **OUTERMOST** in `deliver_to_session` — `interrupt` does NOT override it; a replaced session must never be woken into its successor's conversation
- also guards the **bg-resume choke point** (`build_enqueue_callback`), since background completions resolve their route directly, bypassing `deliver_to_session`
- **fails open** on unknown ownership — the gate never makes a surface unnotifyable
- names the occupying session in the refusal, so a sender can redirect without a discovery round

Ownership is tracked in `TelegramState.channel_ownership`, a sync mirror maintained ONLY at `register_session_chat` (the single write site of the async maps), because the delivery gate's probe is a sync closure that cannot take tokio mutexes.

## Port seam (fork → upstream)

The fork's bg-resume choke additionally **parked** the undeliverable completion (`park_undeliverable` + `wait_ready`). Neither exists upstream — there is no channel-park primitive in `build_enqueue_callback`'s callback. Upstream behaviour here is therefore **warn + drop**: the completion is refused with a loud `tracing::warn!` naming the successor session. Rationale: parking would hold the message forever anyway — the replaced session will never be claimed again. The gate's contract is the wake refusal, not message preservation.

## Tests

8 new tests in `src/tests/session_notify_test.rs`: occupied refuses (with and without interrupt), ownership gate outranks turn gate, owned/unknown ownership pass, tool refusal text names the occupant, mirror tracks channel replacement (incl. DM vs general bucket keys. 3 existing call sites adapted to the 3-arg signature (`session_notify_test.rs` ×2, `channel_route_expectation_test.rs` ×1).

CI: pr-checks GREEN — https://github.com/leshchenko1979/opencrabs/actions/runs/33324521740

Original issue: https://github.com/leshchenko1979/opencrabs/issues/17

### DIFF
--- a/src/brain/agent/service/session_routes.rs
+++ b/src/brain/agent/service/session_routes.rs
@@ -12,7 +12,7 @@
 //! unrelated concerns shared one file and one set of locks to reason about.
 
 use std::collections::HashMap;
-use std::sync::Mutex;
+use std::sync::{Arc, Mutex};
 
 use uuid::Uuid;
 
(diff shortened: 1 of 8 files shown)


leshchenko1979:
CI re-trigger via close/reopen (no code change): the Test step drew the known pre-hardening governor-gates lottery — `governor_gates_test::queued_final_drains_over_the_wire_through_mock_bot_api` panicked at `src/tests/governor_gates_test.rs:310:15` ("...but received 2"), run https://github.com/adolfousier/opencrabs/actions/runs/33331093480 (7267 passed / 1 failed; all channel-ownership tests green). The hardened deterministic test is inbound via #1263. This PR's tree predates that hardening, hence the redraw.

PR #1268, opened 2026-08-31
feat(notify): redirect delivery to the new occupant when a session's channel is claimed (leshchenko1979/opencrabs#19)

leshchenko1979:
## What

Closes the gap left by the refusal-only ownership gate: when a session's channel (DM or forum topic) has been claimed by a newer session, `session_notify` delivery no longer bounces with an error — it is **redirected to the current occupant**, with provenance framing so the occupier's conversation knows where the message came from.

- **Redirect loop with hop cap**: delivery to an occupied session re-routes to the occupant; if the occupant's channel is *also* occupied, the chase continues, capped at 3 hops — on cap the message is parked (logged loudly) instead of looping.
- **Provenance framing**: redirected messages carry a header naming the original target, so the receiving session can contextualize the content.
- **`Delivery::Redirected`** replaces the fork's `RefusedChannelOccupied` shape: both `session_notify` tool and background-task completion callers report the redirect as a success outcome with the new occupant named.
- **DRY reroute**: the background-task completion path (which previously resolved its enqueue route directly, bypassing `deliver_to_session`'s ownership gate) now flows through the single `deliver_to_session` route. One route, one semantics — the duplicate channel-ownership choke in the Telegram `resume.rs` enqueue callback is removed (its only reason to exist was that bypass).
- Folds in the four clippy findings from the follow-up cleanup (collapsed nested-if → let-chain, import cleanup) so the feature lands lint-clean.

## Tests

`session_notify_test.rs` carries the fork suite ported to the upstream tree: mirror keys DM and general buckets separately, mirror tracks channel replacement, redirect reports the new occupant, hop cap parks on multi-hop chains, tool reports redirects as success. The pre-existing refusal tests were converted to their redirect counterparts. Full suite: 14 tests in the file, gate run below.

## Port seams (fork → upstream)

- Fork's `park_undeliverable` / `wait_ready` machinery in `resume.rs` is **not** ported (fork-only lineage, separate concerns); the choke removal here keeps upstream's plain bot fetch.
- Fork's `BgTaskMeta` plumbing does not exist upstream; the DRY reroute passes the plain 3-field `QueuedUserMessage`.
- Upstream `ToolResult` exposes `.output`, so redirect assertions assert on `.output`.

## CI

Fork gate run on this exact head (`ab25391b`, proven in job title): https://github.com/leshchenko1979/opencrabs/actions/runs/33352364236 — GREEN (fmt, clippy, tests).

**Note for reviewers**: this branch is stacked on #1267 (`fix(notify): refuse delivery…`), whose commits are not yet on `main` — the file-level diff below therefore shows #1267 + this PR combined. The reviewable delta for *this* PR is the two commits on top of `af2998218792552bf79563e41854e866949ba5ff`. Merge order: this PR after #1267.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/19

### DIFF
--- a/src/brain/agent/service/background_tasks.rs
+++ b/src/brain/agent/service/background_tasks.rs
@@ -12,7 +12,7 @@ use std::sync::Mutex;
 
 use uuid::Uuid;
 
-use super::types::{MessageEnqueueCallback, PushOrigin, QueuedUserMessage};
+use super::types::{PushOrigin, QueuedUserMessage};
 
 /// Result of a finished background command.
 #[derive(Debug, Clone)]
@@ -33,7 +33,6 @@ pub struct RunningTask {
 
 /// Manages background commands and resumes their sessions on completion.
 pub struct BackgroundTaskManager {
-    enqueue: MessageEnqueueCallback,
     /// In-flight background tasks per session.
     ///
     /// Holds the label and start time, not just a count, because a surface has
@@ -46,9 +45,8 @@ pub struct BackgroundTaskManager {
 use super::detached_status::{self, DetachedFinish};
 
 impl BackgroundTaskManager {
-    pub fn new(enqueue: MessageEnqueueCallback) -> Self {
+    pub fn new() -> Self {
         Self {
-            enqueue,
             running: Mutex::new(HashMap::new()),
         }
     }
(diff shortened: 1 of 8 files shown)


leshchenko1979:
Rebased onto current main (the stacked base of #1267 merged in https://github.com/adolfousier/opencrabs/pull/1267, which left this branch conflicting): cherry-picked the two #19 commits onto `34bd8ff2`, new head `3bfc5328`. Content unchanged — auto-merges only. Gate dispatched on the new head, will report.

PR #1392, opened 2026-09-05
fix: tap-redraw followup keyboard re-render after pick (rich preserve + idx ride)

leshchenko1979:
## What

Tap-redraw for followup suggestion keyboards: after a user taps a suggestion, the picked button is re-rendered (mark_picked_button) instead of leaving the stale full keyboard, with the pick record and rich-body emission kept consistent.

## Details

- `take_pending_followup` now returns the tapped **idx** alongside session/text/host — the tap-redraw rewrite runs in a spawned task outside the arm's scope and needs the index (#67).
- `mark_picked_button` corrects span arithmetic in the non-followup paths and adds the missing `tag_open_len` to the verbatim emit slice, so rich bodies with inline buttons re-emit byte-identically around the redraw.
- Nested if-let collapsed into a let-chain in `mark_picked_button`.
- Verified tap redraw against closed leshchenko1979/opencrabs#71 (rich body + merged-host cases).

## Integration note

Upstream's #1226 G busy-guard re-arm calls `restore_pending_followup` with the entry returned from `take_pending_followup`. This unit's reshape had flattened that return to `(session_id, text, idx, host)`, which would leave the re-arm without an entry to restore. The final commit keeps BOTH behaviors: the full `PendingFollowupEntry` rides in the return tuple and `sid` is derived from it at the tap site. No behavior change for either feature; this commit is an integration seam, not part of the original fork unit.

Fork issue: leshchenko1979/opencrabs#67

PR #1394, opened 2026-09-05
feat: tap-redraw picked suggestion buttons on rich hosts (leshchenko1979/opencrabs#67)

leshchenko1979:
## What

Tap-redraw for follow-up suggestion buttons on rich hosts: after the user taps a suggestion, the rich host body's button rows are rewritten in place instead of vanishing — the picked button flips to `style="success"` with a `✓` prefix and `disabled`, every sibling keeps its style and gains `disabled`. Classic and standalone planes stay on the empty-markup strip (teloxide 0.17.0 has no Bot API 10.3 `style`/`disabled` fields on `InlineKeyboardButton`).

## Details

- `mark_picked_button` transforms the production rich markup shape only (`followup:`-data buttons); non-follow-up markup and row containers pass through byte-for-byte. If a client ever ignored the `disabled` form, a second tap routes to the stale-shell guard — no loops.
- `pick_rewrite` gains a `picked_idx` argument; on rich hosts the pick record rides a mark-picked body instead of a plain append. The leshchenko1979/opencrabs#39 invariant is untouched: the pick record is still baked into the body before any transport arm runs.
- `take_pending_followup` returns the tapped idx alongside the entry/text — the rewrite runs in a spawned task lexically outside the arm that binds `idx`. The idx rides as a third tuple element; upstream's `restore_pending_followup` re-arm keeps receiving the full `PendingFollowupEntry`.
- The final commit folds in the sibling style-attr rendering fix (originally landed against `leshchenko1979/opencrabs#71`): a styled sibling that only gains `disabled` keeps its style so the grayed form is the only rendered state.

## Provenance and attribution

Ported from fork-side work by the same author. Original fork SHAs: `4215c019`, `97a136e1`, `357f2bb9`, `5c3d8b5e`, `95adf054`, plus dependency commit `5e66191b` (fork context `leshchenko1979/opencrabs#71`). The `take_pending_followup` and tap-handler commits are re-expressions against upstream's `(PendingFollowupEntry, String)` take shape and the upstream tap-handler structure; the glue-tier and empty-markup-helper shapes from the fork environment were intentionally not carried.

Supersedes #1392, which carried the fork-only glue tier and dropped upstream's own persistence test; this branch contains only the `leshchenko1979/opencrabs#67` scope.

## Testing

- E0308 stale-set completeness method: the `take_pending_followup` tuple extension initially missed upstream's four existing 2-tuple destructure callers in `telegram_followup_persistence_test.rs` (upstream CI flagged all four); a tree-wide grep for remaining 2-tuple destructures of that call proved the fix set complete before the fixup-squash, so the corrected commit ships with the stale set provably closed, not assumed.
- `telegram_followup_persistence_test.rs` (196 lines) is restored intact on this branch — upstream's own persistence suite runs against the ported tap-redraw flow.
- Evidence runs on this head lineage: 33957170609 (first head), 33959465485 (post-fix head — my files clean, remaining red isolated to pre-existing base `theme_picker.rs`/`messaging.rs` fmt), 33960171977 (final head `ef3a032a`). Base-red repair ownership: `theme_picker.rs` -> #1393/#1395, `messaging.rs` -> carried here as `ef3a032a` (style-commit convention).
- Changed-lines-only comparison of each port commit against its fork original: byte-identical for the four clean picks; the two re-expressions diverge exactly per the documented upstream adaptations above.
- Live smoke on the fork (owner-accepted, rich-plane scope): tapped suggestion re-rendered success+✓+disabled, siblings disabled, pick record intact.
- CI run on this PR head branch is the build/test gate; the head SHA is embedded in the required check job names.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/67

### DIFF
--- a/src/channels/telegram/agent.rs
+++ b/src/channels/telegram/agent.rs
@@ -460,6 +460,7 @@ impl TelegramAgent {
                                                             .as_ref()
                                                             .map(|(full, rich)| (full.as_str(), *rich)),
                                                         picked,
+                                                        idx,
                                                     );
                                                 let outcome: Result<(), String> = match rewrite {
                                                     super::suggest_options::PickRewrite::RichHost(
--- a/src/channels/telegram/agent.rs
+++ b/src/channels/telegram/agent.rs
@@ -363,7 +363,7 @@ impl TelegramAgent {
                                         None
                                     }
                                 };
-                                if let Some((entry, text)) = taken {
+                                if let Some((entry, text, picked_idx)) = taken {
                                     let sid = entry.session_id;
                                     let merged_host = entry.host.clone();
                                     let (chat_id, thread_id, prompt_msg_id) = query
@@ -460,7 +460,7 @@ impl TelegramAgent {
                                                             .as_ref()
                                                             .map(|(full, rich)| (full.as_str(), *rich)),
                                                         picked,
-                                                        idx,
+                                                        picked_idx,
                                                     );
                                                 let outcome: Result<(), String> = match rewrite {
                                                     super::suggest_options::PickRewrite::RichHost(
--- a/src/channels/telegram/state.rs
+++ b/src/channels/telegram/state.rs
@@ -842,10 +842,17 @@ impl TelegramState {
         &self,
         token: &str,
         idx: usize,
-    ) -> Option<(PendingFollowupEntry, String)> {
+    ) -> Option<(PendingFollowupEntry, String, usize)> {
         let entry = self.pending_followups.lock().await.remove(token)?;
         self.forget_followup(token).await;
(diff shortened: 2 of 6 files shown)


leshchenko1979:
Sequencing note for review order: this PR's Lint leg depends on the pre-existing `theme_picker.rs` fmt/clippy base REDs (present at base 3f880c5e, untouched by this branch). Those are repaired by #1393 and #1395 — suggesting reviewing/merging those two first; this branch will rebase onto the new tip immediately after. File coverage is disjoint: #1393/#1395 touch theme_picker.rs only; this PR's style commit (ef3a032a) additionally repairs `messaging.rs:836` fmt, which no other PR covers.

PR #1405, opened 2026-09-05
Fix: parse or strip tool-call JSON returned as text; persist no parser warning into compaction

leshchenko1979:
## Summary

Models that emit tool calls as **text** instead of structured calls could leak raw JSON internals into the conversation: an unparsable residue rode through as visible text, a generic ⚠️ "model does not support function calling" warning was appended as a content block — and that warning then fossilized into every subsequent compaction summary, re-injected into context indefinitely. Observed live on weak/non-streaming models during compaction calls (originally investigated as "agents return garbage"): full report with evidence in leshchenko1979/opencrabs#66.

This PR ports the fix (3 commits, live in production on our fork since 2026-09-01, 4-day smoke: 15 real compaction opportunities, 0 leaks, 0 false positives):

1. **`fix(provider): strip unrecoverable tool-call JSON from responses, flag leak`** — tightens the text detector (parseable call-shaped JSON only; fences excluded — code samples about tool calls no longer trip it), strips the raw JSON at parse time so it can never reach the model loop or a compaction summary, and sets a `tool_text_leak` flag on `LLMResponse` instead of persisting a warning block.

2. **`fix(agent): one corrective retry on tool-text leak, then fail clean`** — one structured-calls nudge retry (mirrors the existing truncation-retry pattern); on a second leak the turn fails with a clean `AgentError` instead of delivering raw internals as the final answer.

3. **`fix(agent): wrap tool-text-leak fail-clean in AgentError::Provider`** — error-path plumbing for the loop's `AgentError` surface.

## Notes

- The `⚠️ does not support function calling` content block is no longer appended to responses, so it can no longer be persisted into compaction summaries (the poisoning vector).
- 38 `LLMResponse` construction sites gain the flag (mechanical, `false` default behavior preserved).
- Rebased onto current `adolfousier/main`; one benign conflict vs the new mermaid-regen retry declarations (both kept).

Environment: fork CI evidence (fmt + clippy `-D warnings` + all-features test) cited in comments after the gate completes.

Refs: leshchenko1979/opencrabs#66

### DIFF
--- a/src/brain/agent/service/helpers.rs
+++ b/src/brain/agent/service/helpers.rs
@@ -1133,6 +1133,14 @@ impl AgentService {
             None
         };
 
+        // Streaming leak flag (fork #66, ex-upstream adolfousier/opencrabs#1260):
+        // text was already emitted to display as deltas, so there is no
+        // retro-strip here — the flag lets the tool loop attempt one
+        // corrective retry and otherwise fail clean instead of accepting
+        // the residue as a final answer.
+        let tool_text_leak =
+            crate::brain::provider::json_repair::content_has_unrecovered_tool_text(&content_blocks);
+
         Ok((
             LLMResponse {
                 id,
@@ -1155,6 +1163,7 @@ impl AgentService {
                     ..Default::default()
                 },
                 streaming_active_secs,
+                tool_text_leak,
             },
             reasoning,
         ))
--- a/src/brain/agent/service/tool_loop.rs
+++ b/src/brain/agent/service/tool_loop.rs
@@ -5548,7 +5548,7 @@ impl AgentService {
                         "provider-integrity",
                         Some("leak persisted after corrective retry; failing clean"),
                     );
-                    return Err(crate::brain::provider::ProviderError::StreamError(msg));
+                    return Err(AgentError::Provider(crate::brain::provider::ProviderError::StreamError(msg)));
                 }
 
                 // ── Mid-sentence truncation retry ────────────────────────
--- a/src/brain/agent/service/tool_loop.rs
+++ b/src/brain/agent/service/tool_loop.rs
@@ -7219,6 +7219,7 @@ impl AgentService {
                     // per-iteration LLMResponses; this top-level
                     // synthesis is for content/usage handoff only.
                     streaming_active_secs: None,
+                    tool_text_leak: false,
                 }
             }
             None => {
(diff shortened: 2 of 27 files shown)


leshchenko1979:
CI evidence (fork carrier, tests the exact PR tree `e074b022c0166460ef9c34da8833c5aa0e10eb13` — sha in job name): https://github.com/leshchenko1979/opencrabs/actions/runs/33993477069 — conclusion **success** (fmt + clippy `-D warnings` + all-features tests).

Live-fire evidence from the fork deployment (4 days, 15 compaction opportunities on the exact trigger path): 0 leaks, 0 false positives — see leshchenko1979/opencrabs#66 for the full investigation and smoke receipts.

adolfousier:
Merged locally and verified. Thanks for this: stripping the JSON at parse time so it can never reach the model loop or a compaction summary is the right shape, and replacing the persisted `⚠️` content block with a flag removes the actual poisoning vector rather than papering over it. The fence exclusion is a good call too.

Full gate on the merge: `clippy --locked --lib --bins --tests --all-features -- -D warnings` clean, `fmt --check` clean, **7,856 tests pass, 0 failures**.

Two things fixed on top rather than sent back, so nothing is blocked.

## 1. A missed construction site

`src/tests/loop_break_test.rs:155` builds an `LLMResponse` and did not get the new field, so the test target failed to compile:

```
error[E0063]: missing field `tool_text_leak` in initializer of `provider::types::LLMResponse`
```

That file almost certainly landed on `main` after your branch was cut, so this is a rebase artifact rather than something you missed. Added `tool_text_leak: false` there.

## 2. Tests for the detector

The PR's fifteen test-file changes are all mechanical field additions; the 173 lines of detection logic and 54 of retry logic arrive unpinned. Given the rules are subtle, I added twelve tests to `brain_provider_json_repair_test.rs`:

- each of the three call shapes is detected
- **fenced JSON is never a leak** — the regression the rewrite exists for
- prose about tool calls with unparseable JSON does not trip it, which is exactly what the old `contains "function" && contains "arguments"` heuristic got wrong
- parseable but non-call-shaped JSON is left alone
- structured `ToolUse` blocks suppress the leak verdict even when the text also contains call-shaped JSON
- stripping removes the leak and preserves the surrounding prose

## One behaviour worth your confirmation

Shape 2 matches on the **key alone**:

```rust
if obj.contains_key("command") {
    return true;
}
```

So any unfenced object carrying a `command` field is treated as a leaked invocation regardless of intent — a model writing `{"command": "npm test"}` inline as illustration gets stripped and draws a corrective retry. The fence exclusion is the escape hatch, and models usually fence, which is presumably why your 4-day smoke saw 0 false positives across 15 compaction opportunities.

I have pinned the current behaviour in a test rather than changed it, since your production evidence is stronger than my reading. Worth confirming it is deliberate: it is the widest of the three shapes, and the test now makes any future narrowing an explicit decision.

Issue #92, opened 2026-05-19
A2A gateway ignores agent approval_policy (auto-always); tools blocked with no approval callback


leshchenko1979:
## Summary

When an external client triggers the agent via the **A2A gateway** (`POST /a2a/v1` `message/send`), tools fail with:

> Tool requires approval but no approval mechanism configured

This happens even when `[agent] approval_policy = "auto-always"` is set in `config.toml` (same profile, same binary v0.3.22).

The **Telegram channel** on the same profile works: tools run and replies surface in chat.

## Expected behavior

With `approval_policy = "auto-always"`, A2A tasks should auto-approve tools the same way headless/daemon paths do when `with_auto_approve_tools(true)` is set — or via `check_approval_policy()` like Telegram's approval callback.

## Actual behavior

A2A `message/send` creates a task and session, but any tool that `requires_approval_for_input` is denied because:

1. `ChannelFactory::create_agent_service()` does **not** call `with_auto_approve_tools()` (defaults to `false`).
2. `a2a/handler/send.rs` `process_task()` calls `send_message_with_tools_and_mode()` with **no** `override_approval_callback`.
3. `tool_loop.rs` hits the branch: needs approval + `approval_callback` is `None` → deny with "no approval mechanism configured".

## Code pointers (v0.3.22)

| Location | What happens |
|----------|----------------|
| `src/cli/ui.rs` | Main agent: `with_auto_approve_tools(config.agent.approval_policy == "auto-always")` |
| `src/channels/factory.rs` `create_agent_service()` | Channel/A2A agent: **no** `with_auto_approve_tools` |
| `src/a2a/handler/send.rs` `process_task()` | Uses `send_message_with_tools_and_mode()` — no approval override |
| `src/channels/telegram/handler.rs` | Uses `send_message_with_tools_and_display(..., Some(approval_cb), ...)` where `make_approval_callback` calls `check_approval_policy()` |
| `src/utils/approval.rs` | `check_approval_policy()` returns `Some((true, true))` for `auto-always` / `auto-session` |
| `src/brain/agent/service/tool_loop.rs` ~3274–3551 | Approval gate + "no approval callback configured" error |

Factory comment says A2A can use "session context" for auto-approval — that does not appear wired in `process_task()`.

## Suggested fix (either is enough; both is fine)

**Option A — factory (matches TUI/daemon):**

```rust
// src/channels/factory.rs — create_agent_service()
[...]

Issue #58, opened 2026-04-07
[Feature] bang operator to directly execute commands


suse-coder:
Currently I cant with ! do directly execute cli commands in the chat.

<img width="939" height="206" alt="Image" src="https://github.com/user-attachments/assets/89c5fc4b-36e8-492e-9213-d21e1828934f" />

PR #1426, opened 2026-09-07
fix(telegram): re-run the #58 promotion guard on the displaced host in options-reclaim (#92)

leshchenko1979:
## What

In `settle_options_reclaim`, when a long demoted host message forces the trailer's promotion to answer (#58 path), the demoted host text is returned raw — if it carries a Telegram code fence, that fence ships unrendered because the fence→PNG conversion runs only on the final-answer send. This PR re-runs the #58 promotion guard on the demoted host before returning it: a fence-bearing (or oversized) displaced host is concatenated into the answer slot so BOTH texts ride the render path. The promoted trailer keeps the answer head (#58 priority stands); the trailer slot dies with nothing left to ride.

## Details

- **Guard re-run (#92):** inside the promotion branch, the demoted host is checked with the same `trailer_promotes_to_answer(should_render_mermaid(d), d.chars().count())` gate the trailer faces; if it trips, `return (format!("{t}\n\n{d}"), None)` — concatenated after the promoted trailer.
- **Regression test:** `settle_demoted_host_trips_guard_concatenates_into_answer` — host (605 chars) and trailer (608 chars) both above `MAX_TRAILER_CHARS`, asserts the promoted trailer keeps the answer head, the tripping demoted host concatenates after it, and the trailer slot is `None`.
- Clean cherry-pick from leshchenko1979/opencrabs (commit `1097b444ab756ddf1f6b43bf547bfe75879120c0`, patch-id verified, trailers `Session-Id` + `Issue-Ref` intact); shipped and smoke-verified in production on the fork (deployed sha `1097b444`, features `telegram,code-graph`).

## Evidence

oc-harvest-sweep CLEAN (trailers + non-empty diff vs `main`). PR-lane gate runs on this head: `34061653423`, `34063416863`, and `34084920421` — fmt + Clippy green, full suite **7859 passed / 1 failed** each; the sole failure is `xiaomi_models_fetch_returns_mimo_list`, the known upstream-base network-dependent onboarding test (no offline xiaomi baseline on this tree; the offline baseline fix is pending in adolfousier/opencrabs#1424). The failure is disjoint from this change (no onboarding/fetch file touched), reproduced identically across three runs with disjoint diffs, and matches the established sole-failure exemption precedent (n=1615; family-C precedent PR #1422). Gate run `34084920421` is terminal: `failure` on the same sole test — final evidence state for this head.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/92

**Update (rebase, 2026-09-07):** rebased onto fresh upstream main `221d7423` (carries the #1424 xiaomi offline baseline). Clean-GREEN gate on the rebased head `3c4ac6c0`: [run 34140891456](https://github.com/leshchenko1979/opencrabs/actions/runs/34140891456) — **full suite pass, 0 failures**, fmt + clippy green. The exempted-RED evidence above is superseded: no exemption is needed for this head.

### DIFF
--- a/src/channels/telegram/flow.rs
+++ b/src/channels/telegram/flow.rs
@@ -2159,6 +2159,21 @@ pub(crate) fn settle_options_reclaim(
         } else {
             Some(text)
         };
+        // #92 demoted-host guard: re-run the promotion guard on the demoted
+        // host. A fence-bearing (or oversized) displaced host riding the
+        // trailer slot would ship its fence raw — the fence→PNG conversion
+        // runs only on the final-answer send (#92). Concatenate it into the
+        // answer slot (owner-approved fix shape) so BOTH texts ride the
+        // render path; the promoted trailer keeps the answer head (#58
+        // priority stands), the trailer slot dies with nothing left to ride.
+        if let Some(d) = demoted.as_ref().filter(|d| {
+            trailer_promotes_to_answer(
+                super::rich::mermaid::should_render_mermaid(d),
+                d.chars().count(),
+            )
+        }) {
+            return (format!("{t}\n\n{d}"), None);
+        }
         return (t.clone(), demoted);
     }
     (text, trailer)
--- a/src/tests/telegram_options_reclaim_test.rs
+++ b/src/tests/telegram_options_reclaim_test.rs
@@ -302,3 +302,34 @@ fn settle_oversized_trailer_empty_host_drops_demote() {
         "empty host demotes to None, not an empty trailer"
     );
 }
+
+#[test]
+fn settle_demoted_host_trips_guard_concatenates_into_answer() {
+    // #92 demoted-host leg (the #58 matrix covered the trailer leg only):
+    // the displaced host itself trips the promotion guard — an oversized
+    // run, the config-independent leg (`should_render_mermaid` reads
+    // `Config::current()`; the fence leg shares the same guard call pinned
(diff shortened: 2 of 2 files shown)


adolfousier:
Merged locally and pushed: the change is in `main` as of `7738b765`, and `src/channels/telegram/flow.rs` on `main` is byte-identical to this branch.

GitHub could not close this automatically because the branch was squashed and force-pushed after the merge, so the head SHA here (`3c4ac6c0`) is not the one that landed (`22f588a4`). Closing by hand for that reason, not because anything was dropped.

Reviewed as merged, no gaps to fill. Re-running the `#58` promotion guard on the demoted host is the right shape: it reuses `trailer_promotes_to_answer` rather than restating the predicate, so the fence and oversize thresholds cannot drift apart from the promotion path they mirror. Concatenating into the answer slot means both texts ride the render path, the promoted trailer keeps the answer head so the `#58` priority still holds, and the trailer slot dies with nothing left to ride.

Your regression test covers all three of those, which is what made this quick to verify.

Full gate green on the merge commit: `cargo fmt --check` clean, `cargo clippy --locked --lib --bins --tests --all-features -- -D warnings` clean, 7924 tests passed, 0 failed.

Useful to know going forward: #1424 has landed, so `xiaomi_models_fetch_returns_mimo_list` passes on a current base. A failure there is a real signal now rather than an expected one.

Issue #814, opened 2026-07-26
REGRESSION (3c45e41a): plan card re-stick floods Telegram and posts duplicates


adolfousier:
## Symptom

A group chat accumulated four identical plan cards within two minutes (21:36, 21:36, 21:37, 21:38), each with its own "Discard plan" button, all showing the same unchecked checklist. No restart occurred, so this is not the tracking-loss path fixed in #809.

## Evidence

`~/.opencrabs/logs/opencrabs.2026-07-26`, one line every ~1.7 seconds:

```
21:37:24  Telegram plan card create failed: Retry after 40s
21:37:25  Telegram plan card create failed: Retry after 39s
21:37:27  Telegram plan card create failed: Retry after 37s
...
21:38:04  Telegram plan card create failed: Retry after 3s
```

Roughly twenty create attempts across forty seconds while Telegram counts down telling the bot to stop. The countdown decreasing in step with wall-clock shows the same flood-control window being hit again and again, not separate incidents.

## Mechanism, three faults compounding

**1. The re-stick is two writes per settle.** `handler.rs:3961` deliberately calls `remove_plan_card` then `refresh_plan_card`, to move the card to the bottom rather than edit a buried message. That is a delete plus a create every turn, on top of the refreshes already firing from the streaming path (`handler.rs:3509`) and the resume path (`resume.rs:297`). On a busy session that is a lot of writes for chrome.

**2. `Retry after N` is ignored.** The create error is logged and dropped (`plan_card.rs:184`). Nothing records that the API asked for a pause, so the next refresh attempts another write immediately. Each rejected attempt keeps the window alive, which is why the countdown never actually elapses.

**3. A failed create loses tracking permanently.** `set_plan_card` only runs on success, so after a failure there is no tracked id at all. The next refresh therefore cannot edit — its only option is another create, which fails the same way. The failure mode feeds itself.

The visible duplicates are the creates that DID land between rejections, while the corresponding deletes did not.

## Fix direction

- **Honour the retry window.** Parse `Retry after N` and suppress card writes for that session until it passes. This is chrome; skipping it is strictly better than being throttled into a loop.
- **Stop re-sticking unconditionally.** Moving the card to the bottom every settle is a nice-to-have that costs two writes per turn. It should be rate-limited, or skipped when the card is already recent, or dropped in favour of editing in place.
- **Do not lose tracking on a failed write.** A create that fails should leave the session in a state where the next attempt can still edit an existing card rather than create another.

## Note

[...]

Issue #69, opened 2026-04-09
Copy paste causes crash


MELSAID888:
I have tried multiple times to copy and paste text from Chrome console. I noticed if the text has urls and special characters, the rust application crashes. 

I beleive pasting need to be wrapped like in other tools where it says (pasted 1-4 lines) to avoid the crash.

adolfousier:
Thanks for reporting this, replicating the issue and pushing a today asap

PR #1427, opened 2026-09-07
Restick plan card on interrupt/teardown turns (fork #69)

leshchenko1979:
## Summary

Interrupted/replaced streaming turns (cancel-token teardown path) early-returned before the plan-card restick block ran, so the plan card could go stale after an interrupt — the restick guarantee held only for normal settles.

**Fix:** the post-turn restick tail is extracted verbatim into `plan_card::restick_plan_card_after_turn` (finalize-on-archive → tracked-card claim on the shared sticky budget → G2 edit / G3 post). The normal-settle path forwards its already-computed keyboard (bit-identical behavior); the interrupt teardown now calls the same tail before its early return, with the keyboard recomputed for the finished turn. No new throttle; the #814 cooldown stays deleted; sticky governors untouched.

## Tests

New `telegram_plan_card_interrupt_restick_test` module pins: shared sticky-budget refusal across both paths, cardless turns spend nothing, keyboard recompute seam for finished turns. No network, no Bot dependency.

## Evidence

- Gate: [PR-lane gates (6db2f6f5)](https://github.com/leshchenko1979/opencrabs/actions/runs/34100393773) — fmt + clippy green; tests 7860 passed / 1 failed
- The single failure is `xiaomi_models_fetch_returns_mimo_list`, a pre-existing upstream-base defect: the test requires live internet to api.xiaomimimo.com and the xiaomi branch of `fetch_provider_models` has no offline baseline (upstream diagnosis verified from source; independent runs on disjoint diffs fail identically). This PR's diff does not touch onboarding. Precedent for merging with this sole failure: #1422, #1426. A pending upstream fix exists as #1424.
- Drift disclosure: cherry-pick from the fork dropped one rustfmt-only reflow hunk in `refresh_plan_card` (base drift, upstream's newer goal-section load shape kept); the substantive helper is byte-identical between fork and this head, verified by diff.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/69

**Update (rebase, 2026-09-07):** rebased onto fresh upstream main `221d7423` (carries the #1424 xiaomi offline baseline). Clean-GREEN gate on the rebased head `45f35190`: [run 34140966104](https://github.com/leshchenko1979/opencrabs/actions/runs/34140966104) — **full suite pass, 0 failures**, fmt + clippy green. The exempted-RED evidence above is superseded: no exemption is needed for this head.

### DIFF
--- a/src/channels/telegram/handler.rs
+++ b/src/channels/telegram/handler.rs
@@ -2622,6 +2622,26 @@ pub(crate) async fn handle_message(
         if let Some(mid) = streaming_msg_id {
             best_effort_delete(&bot, msg.chat.id, mid, "keep-history teardown").await;
         }
+        // #69: a preempted turn used to `return Ok(())` here and skip the #62
+        // re-stick entirely — during an interrupt chain (rapid-fire messages,
+        // exactly the scenario #62 targets) every reply was interrupt-delivered
+        // and the plan card stayed buried at a stale position until a normal
+        // settle finally happened. Run the SAME settle-path tail here: the
+        // keyboard is recomputed for a finished turn (turn_active == false, so
+        // Approve/Discard rides the resticked card), and the shared sticky
+        // budget (#1150) + governors G2/G3 gate it exactly like a normal
+        // settle — no new throttle, no #814 cooldown reintroduced.
+        let (_, plan_kb) = super::flow_chrome::load_plan_state_section(session_id, false).await;
+        super::plan_card::restick_plan_card_after_turn(
+            &bot,
+            msg.chat.id,
+            thread_id,
+            &telegram_state,
+            &agent,
+            session_id,
+            plan_kb,
+        )
+        .await;
         return Ok(());
     }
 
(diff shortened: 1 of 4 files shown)


adolfousier:
Merged locally and pushed: the change is in `main` as of `b14a9111`, and both `src/channels/telegram/plan_card.rs` and `src/channels/telegram/handler.rs` on `main` are byte-identical to this branch.

GitHub could not close this automatically because the branch was squashed and force-pushed after the merge, so the head SHA here (`45f35190`) is not the one that landed (`3f0593da`). Closing by hand for that reason, not because anything was dropped.

Reviewed as merged, no gaps to fill. The extraction is faithful in both directions: the normal settle path forwards the keyboard value its own render just computed, so that path is unchanged, and the interrupt teardown recomputes it for a finished turn so Approve/Discard rides the resticked card. Gating is identical on both, finalize-on-archive first, then the tracked-card claim on the shared sticky budget, then refresh riding G2 and fresh post riding G3. No new throttle, and the `#814` cooldown stays deleted, which the new doc comment records so the next reader does not reintroduce it.

Catching that a preempted turn returned before the re-stick is the good part of this: an interrupt chain is precisely the scenario the re-stick exists for, so the card stayed buried exactly when it mattered most.

Full gate green on the merge commit: `cargo fmt --check` clean, `cargo clippy --locked --lib --bins --tests --all-features -- -D warnings` clean, 7926 tests passed, 0 failed. The rustfmt-only hunk this description disclosed as dropped from the cherry-pick did not surface here; the formatting drift that did turn up was in the parent-binding branch and is fixed on `main` in `e990ff57`.

Useful to know going forward: #1424 has landed, so `xiaomi_models_fetch_returns_mimo_list` passes on a current base. A failure there is a real signal now rather than an expected one.

Issue #1419, opened 2026-09-06
onboarding: xiaomi /models fetch has no static fallback — network-coupled test fails on 401


leshchenko1979:
## Problem

`tests::xiaomi_onboarding_test::xiaomi_models_fetch_returns_mimo_list` (`src/tests/xiaomi_onboarding_test.rs`, assertion at :103) calls `fetch_provider_models` for the Xiaomi provider, which GETs the keyless live endpoint `https://api.xiaomimimo.com/v1/models` (`src/tui/onboarding/fetch.rs`, xiaomi branch at :330). When the endpoint answers non-200 — consistently **HTTP 401** on our fork's CI runners while upstream runners receive 200 for the identical code — the function returns an empty list and the test panics: `/models must show a model list for Xiaomi (was empty)`. The test is therefore coupled to a live third-party endpoint whose behaviour is environment/geo-dependent.

The MiniMax branch of the same function already solves this exact case: on fetch failure it falls back to a binary baseline list (`merge_minimax_baseline(minimax_baseline_models(), user_models)`, precedent at :288). The Xiaomi branch has **no** such fallback — so the onboarding model picker is empty whenever the live call fails, and the test red-lines the whole Test job.

## Evidence

- Sole failure on every Test run of stacked PRs #1414–#1418, e.g. run [34019518072](https://github.com/adolfousier/opencrabs/actions/runs/34019518072) (head `3469aafb`): 7859 passed, 1 failed — log line `[fetch_provider_models] xiaomi /models HTTP 401 Unauthorized` (fmt at fetch.rs:368).
- Same code green on upstream main in the same window (run 34009441993 at tip `37726ae1`, cited in the fork-side tracking issue) → environment difference, not a code defect, but deterministic RED on runners that receive 401.
- Fork-side tracking obligation: leshchenko1979/opencrabs#112.

## Fix direction

Mirror the MiniMax precedent: a `xiaomi_baseline_models()` static list merged via `merge_minimax_baseline` whenever the live fetch returns empty. The picker is never empty on failure, and the test passes with no dependency on the live endpoint (acceptance per leshchenko1979/opencrabs#112: `cargo test xiaomi_onboarding_test` passes with no network access).

adolfousier:
Fixed on `main` in 4ed11415, and now pinned by regression tests in e9625e40.

**Root cause.** The xiaomi branch of `fetch_provider_models` GETs a live keyless
[...]

Issue #26, opened 2026-03-05
Reply to messsage in Telegram


opryshok:
Opencrabs doesn’t pick up when I reply to a specific message. For example, if it sent me something 10 messages ago that I want to reference now, I reply to that message, but it doesn’t get the context.

PR #1434, opened 2026-09-07
Bind sub-agent status to the spawning session — deliver revived agent outcomes to the parent

leshchenko1979:
### Problem

Upstream (as of v0.5.0) has a gap in sub-agent restart recovery:

1. `spawn.rs` writes the **child's own session id** into `WorkStatus.session_id`, while `restart_recovery.rs` routes the boot-time interruption report to that same field **assuming it is the parent** ("routing interruption reports by the pre-#26 parent_session_id value" per its own comments). The report therefore goes to the dead child session, not to the parent that is waiting for the result.
2. If the child's session *is* resumed by boot-resume and completes, the completion outcome is routed to the recorded surface (default `tui`) — nobody receives it. The parent never learns the agent finished, and the status file is left non-terminal.

Observed live on a production fork deployment: a revived sub-agent finished its work, the parent received nothing, and the status file zombied at `Running`.

### Fix (mirrors a fix proven in production on a fork deployment)

- `work_status.rs`: `WorkStatus` gains `parent_session_id: Option<String>`, stamped at spawn (before the first status write). New helper `find_agent_by_session` finds a non-outcome-terminal agent (`Pending`/`Running`/`Interrupted`/`AwaitingInput`) by its session id — `Interrupted` is included because reconcile marks the file at boot *before* the resumed child completes.
- `spawn.rs`: passes the spawning session at both construction sites.
- `restart_recovery.rs`: interruption reports route to the parent when the child session carries a parent binding (falls back to prior behavior when absent); new `deliver_revived_agent_outcome` delivers the resumed child's outcome to the parent and finalizes the status file.
- `cli/ui.rs`: both resume-completion arms check `find_agent_by_session` and route via `deliver_revived_agent_outcome`, skipping the phantom-surface path.
- `work_status_parent_binding_test.rs`: 5 tests (parent round-trip, lookup hit/miss/skip, legacy files without the field parse as `None`).

`parent_session_id` is `Option` + `skip_serializing_if`, so existing on-disk status files parse unchanged.

### Rebase note (supersedes #1431)

Refiled after upstream main moved to `221d7423` (#1424/#1422 merges). Rebased cleanly; the earlier xiaomi baseline-fallback rider is **OBSOLETE-DROPPED** — upstream landed its own #1419 fix, so carrying ours would duplicate it. Diff now contains only the parent-binding change plus its required test call-site updates.

### Verification

- Fork CI on this exact tree: run 34142281316 (job name carries the full tested sha `25d90c1b…`); green-run confirmation comment follows the gate verdict.
- Live-fire proof on the fork deployment: sacrificial sub-agent spawned → daemon bounced mid-run → boot-resume revived the child → **parent received the child's result verbatim**, status file finalized (`Completed` + `completed_at`). Same scenario pre-fix: result lost to the `tui` surface, status zombied.

Related fork issues: [leshchenko1979/opencrabs#110](https://github.com/leshchenko1979/opencrabs/issues/110) (defect + live-fire evidence), [leshchenko1979/opencrabs#113](https://github.com/leshchenko1979/opencrabs/issues/113) (v0.5.0 delta analysis).

### DIFF
--- a/src/brain/agent/service/restart_recovery.rs
+++ b/src/brain/agent/service/restart_recovery.rs
@@ -481,7 +481,7 @@ pub(crate) fn deliver_revived_agent_outcome(
     let agent_id = status.id.clone();
     let label = status.label.clone();
     let msg = crate::brain::tools::subagent::spawn::completion_message(&label, &agent_id, outcome);
-    match crate::brain::agent::service::session_routes::deliver_to_session(parent, msg, true) {
+    match crate::brain::agent::service::session_routes::deliver_to_session(parent, msg.clone(), true) {
         crate::brain::agent::service::session_routes::Delivery::Delivered => {
             tracing::info!(
                 target: "background_task",
--- a/src/brain/agent/service/restart_recovery.rs
+++ b/src/brain/agent/service/restart_recovery.rs
@@ -361,22 +361,25 @@ pub async fn recover(local: Option<MessageEnqueueCallback>) -> usize {
     // not, so every file still mid-flight is an agent that no longer exists.
     let orphans = crate::brain::tools::subagent::reconcile::reconcile_orphaned_agents();
     let mut reported = 0usize;
-    for orphan in orphans {
-        match Uuid::parse_str(&orphan.session_id) {
-            Ok(session_id) => {
+    for mut orphan in orphans {
+        match interrupted_report_target(&orphan) {
+            Some(session_id) => {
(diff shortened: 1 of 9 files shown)


leshchenko1979:
Recreating — head ref resolved to a nonexistent branch on the fork (doubled-owner prefix), which pinned a stale sha. Superseded by the incoming PR.

Issue #219, opened 2026-06-23
Skill body lost after compaction — no reinjection mechanism like tools/plans


leshchenko1979:
## Problem

After a skill is invoked (`/repo-audit`, `/security-audit`, etc.), the skill body is injected as a **user message** in the conversation history. The system prompt only carries a compact **index** (name + 120-char description) via `push_commands_and_skills()`.

After context compaction (or enough turns to push the skill body out of the window), the model **forgets the detailed workflow instructions**. It still sees the index (`- /repo-audit: Run a comprehensive language-agnostic repository health audit...`) but no longer has the actual steps, templates, or reference material from the SKILL.md body.

This causes:
- Agent stops following the skill workflow mid-execution
- Agent improvises instead of following the prescribed steps
- Multi-step skills that span many turns degrade after ~5-10 turns

## How tools and plans handle this (skills don't)

| Mechanism | Reinjection after compaction | Session-scoped activation |
|---|---|---|
| **Tools** | JIT activation in `execute()` promotes to `active_extended`; schemas re-sent every request via `tool_schemas_for_session()` | ✅ `registry.activate_tools(session_id, names)` |
| **Plans** | `plan start` (no args) re-surfaces in-progress task details; plan JSON persisted to disk | ✅ per-session `.opencrabs_plan_<session-id>.json` |
| **Skills** | ❌ Nothing. Body injected once as user message, then lost | ❌ No session tracking |

The BRAIN_PREAMBLE has reminders for tools ("PROACTIVE TOOL DISCOVERY — call `tool_search`...") and plans ("`start` with no args to re-surface..."). Skills have **no equivalent reminder**.

## Root cause

In `prompt_builder.rs`, `push_commands_and_skills()` only injects the skill **index**:

```
--- Available Commands & Skills ---
Skills — saved workflows triggered by `<slash>`:
- /repo-audit: Run a comprehensive language-agnostic repository health audit...
- /security-audit: Run a comprehensive language-agnostic security & CVE audit...
```

The full body only appears when `match_user_command()` in `commands.rs` resolves the skill and forwards `skill.body` as a `UserPrompt`. Once that message is compacted away, it's gone.

### Option A: Active skill tracking (mirrors tool activation)

1. Add a per-session `active_skills: HashSet<String>` to the agent service (similar to `active_extended` for tools)
2. When a skill is invoked, add it to `active_skills`
[...]

PR #1438, opened 2026-09-07
feat(compaction): advisory skill-inventory stamp on post-compaction prompts

leshchenko1979:
## Summary

Post-compaction continuation prompts now carry an **advisory skill-inventory stamp**: `SKILLS LOADED PRE-COMPACTION: <names>` + "consider whether each is still relevant; reload only those that are". Advisory, not prescriptive — session focus may have shifted after a compaction.

- Recording uses the existing `AgentService.active_skills` registry (#219) — peek, no new state; the map is never cleared so #219 body re-injection keeps working
- All 5 compaction kinds (Regular/MidLoop/Emergency/PostTool/Manual) × fun/silent variants
- Zero marginal LLM tokens: composed mechanically at prompt-build time; no stamp when zero skills were invoked
- Second commit: DRY — the five verbatim `build_continuation` call sites collapse into one `continuation_prompt()` helper (net −9 lines)

Original issue: leshchenko1979/opencrabs#125

## Test evidence

- Fork triad gate (fmt+clippy+tests, 7275 tests): https://github.com/leshchenko1979/opencrabs/actions/runs/34138907147 — GREEN on 5e0b7080 (identical diff to this branch's 6cfae040; upstream base was re-verified fresh 221d7423 at harvest time)
- 4 new unit tests: stamp framing, empty-set silence, all-kinds presence, sort determinism
- Live smoke: deployed binary 5e0b7080 — `SKILLS LOADED PRE-COMPACTION` string present in /proc exe, disk==proc verified

### DIFF
--- a/src/brain/agent/service/compaction_prompts.rs
+++ b/src/brain/agent/service/compaction_prompts.rs
@@ -141,6 +141,33 @@ pub fn build_continuation(
     text
 }
 
+/// Advisory skill-inventory stamp (#125): names the skills that were
+/// active (slash-invoked) in this session before the compaction, framed
+/// as CONSIDER-not-RELOAD — the session's focus may have shifted while
+/// the context was compacting, so the waking agent weighs each skill's
+/// relevance against the IMMEDIATE TASK instead of blindly re-reading
+/// bodies it may no longer need. An empty slice produces NO stamp at
+/// all: zero marginal tokens for sessions that never touched a skill.
+///
+/// The list rides the continuation prompt (mechanically composed from
+/// in-memory session state at build time), so it is immune to the
+/// summarizer's token-budget drops — the durable-state property the
+/// research phase identified as the gap.
+pub fn append_skill_stamp(mut text: String, active_skills: &[String]) -> String {
+    if active_skills.is_empty() {
+        return text;
+    }
+    let mut names: Vec<&str> = active_skills.iter().map(|s| s.as_str()).collect();
+    names.sort();
+    text.push_str(&format!(
+        "\n\nSKILLS LOADED PRE-COMPACTION: {}. Session focus may have \
+         shifted — consider whether each is still relevant to the \
+         IMMEDIATE TASK; reload only those that are.",
+        names.join(", ")
+    ));
+    text
+}
+
 fn fun_body(kind: CompactionKind) -> &'static str {
     match kind {
         CompactionKind::Regular => {
--- a/src/brain/agent/service/tool_loop.rs
+++ b/src/brain/agent/service/tool_loop.rs
@@ -1195,11 +1195,17 @@ impl AgentService {
                     // auto-compaction behavior but uses a short sentence instead
                     // of the full POST-COMPACTION PROTOCOL. Persisted to DB so
                     // the next turn sees it.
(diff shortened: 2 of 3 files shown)


adolfousier:
Merged locally on `6cfae040` and reviewed. Clean merge, no conflicts, and the full gate is green here on a current base: `cargo fmt --check` clean, `cargo clippy --locked --lib --bins --tests --all-features -- -D warnings` clean, 7934 tests passed, 0 failed. Whatever CI is showing, it is not this diff against current `main`.

The design holds up. Composing the stamp mechanically at prompt-build time from in-memory session state is the right call, because it puts the inventory outside anything the summarizer can drop under budget pressure, which is exactly the failure a summarized skill list would have. Advisory framing is also right: a compaction is often where focus shifts, so prescribing a reload would be wrong more often than it is helpful. Sorting the names for determinism and emitting nothing at all for a session that never touched a skill are both good instincts.

Collapsing the five verbatim call sites into `continuation_prompt()` is the part I would have asked for if it had not been in the second commit. Five copies of the same four-argument construction is five places to forget.

## One gap, filled on top in `61bf9eb5`

Nothing tested the wiring. The four new tests drive `append_skill_stamp` directly, which is the right way to test the stamp itself, but it means a site restored to calling `build_continuation` would build a perfectly valid prompt with the stamp silently absent, and all four would keep passing. The refactor is what makes the feature correct, and it was the untested part.

Added `src/tests/compaction_single_continuation_path_test.rs`, a source guard pinning the structural property: the loop reaches `build_continuation` exactly once, that path is `continuation_prompt`, and it still both appends the stamp and feeds it from `active_skills_for_session`. A sixth compaction site now has to go through the same door rather than quietly skipping what the other five carry. It applies to whatever rides the continuation next, not just this stamp.

Worth noting the count assertion is not vacuous: it finds exactly one occurrence today, so it proves both that the pattern matches and that a second direct call trips it.

Not closing this yet. It is merged in a local branch and closes when that reaches `main`.

adolfousier (approved):
Approved. Merged locally on `6cfae040`, clean merge, and the full gate is green here on a current base: `cargo fmt --check` clean, `cargo clippy --locked --lib --bins --tests --all-features -- -D warnings` clean, 7934 tests passed, 0 failed.

The design holds up. Composing the stamp mechanically at prompt-build time from in-memory session state puts the inventory outside anything the summarizer can drop under budget pressure, which is exactly the failure a summarized skill list would have had. Advisory framing is the right call too, since a compaction is often where focus shifts. Sorted names and no stamp at all for a skill-free session are both good instincts.

Collapsing the five verbatim call sites into `continuation_prompt()` is the part that makes the feature correct rather than merely present.

One gap, already filled in `61bf9eb5`: nothing covered the wiring. The four new tests drive `append_skill_stamp` directly, which is right for the stamp itself, but it means a site restored to calling `build_continuation` would build a valid prompt with the stamp silently absent and all four would keep passing. Added a source guard pinning that the loop reaches the builder exactly once, that the path is `continuation_prompt`, and that it still feeds from `active_skills_for_session`. It covers whatever rides the continuation next, not just this stamp. Nothing for you to do. 

Thank you very much.

PR #1439, opened 2026-09-07
fix(telegram_send): same-origin-only thread inheritance + landing echo in success output

leshchenko1979:
## What

Two changes to the `telegram_send` tool, fixing the failure mode behind repeated `Bad Request: message thread not found` sends and closing the delivery feedback loop:

**1. Same-origin-only thread inheritance.** When a call omits `thread_id` (and `chat_id`), the resolver previously fell back to the session's last-seen topic *regardless of target chat*. In cron/restarted/compacted sessions that memory is stale, so sends targeting a different chat inherited a dead topic id from the origin chat and Telegram rejected the whole send. Now the session-origin topic is inherited **only when the resolved target chat equals the origin chat**; a different target chat falls through to plain topic lookup, and explicit `thread_id` overrides still win in every case.

**2. Landing echo in success output.** Every message-creating success path (send, reply, forward, poll, buttons, photo, document, location) now appends `Landed: chat <id>, topic <id or name>` — resolved destination, not the requested one. Models previously saw "sent" with no landing info, so misrouting (stale-topic auto-routing, session-origin fallback) was invisible until a human complained. The echo makes the outcome observable to the calling model, enabling self-correction. Degrades gracefully to numeric topic id when the topic-name DB is unavailable.

## Implementation notes

- `resolve_thread_id`: origin-topic inheritance gated on `target == origin_chat`; `thread_id: null` (General) semantics preserved.
- `landing_echo()` helper + `thread_name()` (topic-name lookup, `Option`-degrading) applied to all 8 message-creating success sites.
- Tests in `src/tests/telegram_send_origin_thread_echo_test.rs` (7 cases): same-origin inheritance, no cross-chat leak, explicit override precedence, General fallthrough, echo with/without topic-name DB.
- Verified on the live fleet build: omitted `chat_id`+`thread_id` landed in the calling session's origin topic, success output echoed `Landed: chat …, topic …`.

## Evidence

- Fork CI triad GREEN on this head: https://github.com/leshchenko1979/opencrabs/actions/runs/34154329736 (fmt / clippy `-D warnings` / tests, all success, job name pins head sha `4c7dfb3f`).
- Live smoke on the running box: binary identity match vs build of the fork ship run, behavioral probe PASS (landing echo present, same-origin routing correct).

Original issue: https://github.com/leshchenko1979/opencrabs/issues/127


adolfousier:
Landed on `main` as **`1f684337`**.

PR #1449, opened 2026-09-08
fix(telegram): normalize mermaid render-cache key across delivery surfaces

leshchenko1979:
## Summary
Mermaid diagrams rendered through different delivery surfaces (rich plan-card prose vs raw-fenced telegram_send messages) produced cache keys 1 byte apart (trailing newline handling differs between the two resolve paths), causing a double render of identical diagram content within the TTL window.

This change normalizes the cache key across delivery surfaces, so the same diagram content rendered through any plane hits the same cache entry.

## Evidence
- Behavioral probe on the live box (binary identity: exe sha256 == CI artifact run 33987053203, swapped 2026-09-05 20:53:59Z):
  - Parser plane: MISS → mermaid.ink render ok (bytes=21372) → cache_put under normalized key
  - Raw-fence plane (same content): ZERO render lines — silent cache HIT, fence replaced by image ref, photo delivered. Pre-fix this leg double-rendered.
  - Both legs within TTL window; leg-1 MISS proves fresh in-window creation.
- Full probe write-up: leshchenko1979/opencrabs#100 (comment 5588201947)

Original-issue: leshchenko1979/opencrabs#100


adolfousier:
Landed on `main` as **`3141ac5b`**.

Issue #104, opened 2026-05-21
Telegram footer


leshchenko1979:
Have a single line at the bottom of the agent text message, detailing the context used.

## Rationale
My coding sessions start hallucinating heavily at some point and I would like to understand what exactly bloats context.

## Reference 
Hermes has the same feature - it's concise and non-intrusive.

## Implementation
Something like what TUI has now: "ctx 8K/200K".

Probably, will need to take care of:
- The footer not persisting in the session log
- A config parameter to turn this on/off

adolfousier:
I like this very much, great feature. Implementing it as we speak, dropping into v0.3.25 today.

Did some improvements to context handling, tokens usage and compaction summary prompt, next release should notice improvement in the memory and continuation after an automatic compaction.

adolfousier:
Fixed this. The ctx budget footer was being appended to the response text, which made it show up in the middle of the conversation when there were tool calls or intermediate messages.

Now the footer is sent as a separate message after a 2-second delay, so it always appears at the very end after the complete response is delivered. This applies to all channels: Telegram, Discord, Slack, and WhatsApp.

Commit: b11fe2be

leshchenko1979:
No, it WAS A SEPARATE message before.

adolfousier:
> No, it WAS A SEPARATE message before.

Its always a separated message yes, in your screenshot in the other issue it appears between tool calls and response , this fix just ensure it appears after the final completion was sent.

<img width="499" height="494" alt="Image" src="https://github.com/user-attachments/assets/c4d4b72a-b0cc-47dc-b2a9-44751469e551" />

PR #1451, opened 2026-09-08
feat(scheduler): opt-in forum-thread delivery for telegram deliver_to (#104)

leshchenko1979:
## Summary
- `deliver_to` on cron jobs can carry a Telegram forum thread: `telegram:<chat_id>:<thread_id>` routes the delivery into that topic instead of General.
- Thread-missing / thread-not-found errors fail loudly instead of silently re-routing, so misconfigured crons are visible at fire time.
- Includes follow-up rebase fixes (Requester import, closure var rename) and a clippy cleanup in the thread-missing error arm.

Original fork work: leshchenko1979/opencrabs#104 (closed 2026-09-07, live-verified on the ops box: probe message landed in the target thread 7198; deployed binary carried the fix at ship time).

Smoke: behavioral probe GREEN on the live box (thread 7198 landing receipt, channel_messages DB row verified 2026-09-07). Owner ordered restart + probe re-fire; daemon confirmed on the fixed binary.


adolfousier:
Landed on `main` as **`46fc8fc3`**, **`01a02219`**, **`6efa38e0`**.

Issue #131, opened 2026-05-29
In Telegram, crabs don't see what part of the replied message the user has quoted


leshchenko1979:
<img width="509" height="564" alt="Image" src="https://github.com/user-attachments/assets/06cf1ecc-3639-4542-8d09-29d764250cf5" />

adolfousier:
Fixed in `93ee4004` on main.

When you reply to a message in Telegram and highlight a specific part with quote-reply, the bot now sees what you highlighted, not just the whole message.

The handler reads `msg.quote()` (the new field in teloxide 0.13). Three cases:
- You highlighted part of the message: the prompt now shows both the highlighted text AND the full message, so the agent knows which part you mean AND has the surrounding context.
- You highlighted the whole message (same as full text): shows it once.
- No quote (regular reply): same as before.

11 tests added in `src/tests/telegram_quote_reply_test.rs` covering each path, unicode, multiline, whitespace edges, and bot replies.

adolfousier:
Its working as expected now, coming on v0.3.31 later today

<img width="1280" height="409" alt="Image" src="https://github.com/user-attachments/assets/41f926ae-4e9f-486a-8a25-430cbe803869" />

leshchenko1979:
Perfect!

PR #1454, opened 2026-09-08
feat(skills): load_brain_file slug form + seen-skills registry for compaction stamp (#131)

leshchenko1979:
## Summary
- load_brain_file accepts bare slug form; seen-skills registry stamps consumed skills into the compaction continuation prompt (PostTool + MidLoop), so identity survives compaction without always-inject.

## Smoke evidence (behavioral, v0.4.104 rubric)
Live-box probe on deployed 0c281e51: planted skill read via read_file → PostTool + MidLoop stamps non-empty: `skill inventory stamp = ["probe131"] (active 1/0 + seen 1/1)` (tool_loop.rs:481). Probe receipt verified first-hand by Triage in the oc134probe profile log; both stamp planes captured.

## Original issue
leshchenko1979/opencrabs#131


adolfousier:
Landed on `main` as **`94d99349`**, with follow-ups in **`33133df4`** and **`b1bcb310`**.

**Why cherry-picked rather than merged:** this branch was cut before the `audit <audit@local>` identity rewrite on `main`, so it still carried pre-rewrite twins of ~19 commits already upstream. Merging would have re-imported those old-identity commits and undone the rewrite. `git cherry` isolated the genuinely-new commit by patch-id and it was replayed with your authorship preserved.

**Fixed on top (the port did not build against current `main`):**

- `load_brain_file.rs`: the skill-slug block sits above the `let query` binding it reads, so the tree failed with `E0425: cannot find value 'query'` four times. The binding is hoisted above the block. The nested `if` inside it also tripped `clippy::collapsible_if` under `-D warnings`; folded into a let-chain.
- `seen_skills_test.rs`: `.execute(...)` without the `Tool` trait in scope, `E0599` five times. Import added.
- `unknown_slug_falls_through_to_brain_file_error` asserted a hard error. That was `main`'s behaviour when this was written; `main` now answers a missing brain file with a success-carrying "not found" message. The test keeps the intent that matters — an unresolvable slug must never be served as a skill body — and pins it by content rather than by the success flag.
- The four unit tests arrived as an inline `mod tests` in `src/brain/tools/seen_skills.rs`. Tests live under `src/tests/` here, and the feature already had a file there, so they moved in beside the rest (`b1bcb310`); assertions unchanged.

**Gates on the merged result:** fmt exit 0, clippy `-D warnings` exit 0, `cargo test --all-features` 8,006 passed / 0 failed / 30 ignored. Pushed in `c4621f5d..bf9bf5c9`.

Closing as landed — GitHub cannot auto-detect it because the replayed commits carry new SHAs. Good feature; the slug form is a genuinely better reload surface than the filename form.

Issue #111, opened 2026-05-22
RSI Writes Raw Failure Events to TOOLS.md Instead of Derived Rules


leshchenko1979:
# Suggestion: Move RSI Failure Event Logs Out of TOOLS.md

**Date:** 2026-05-22  
**Status:** Open  
**Category:** RSI / Brain File Bloat  
**Severity:** Low (correctness issue, not functional)

---

## Problem

TOOLS.md has become a **failure event audit log** rather than an operational guidance file. It contains ~30+ sections documenting individual tool failure events with timestamps, session IDs, and counters — accumulated entry-by-entry by the RSI process since May 17th.

### Examples

```
### Timeout Handling (5 failures: 17:02, 16:59, 16:58, 16:57, 16:55)
### Bash Timeouts (3 failures: 5d479d92@04:01, 22:06, 975ec69d@14:24)
## Bash Exit Code 127 — Recurring (6 failures since 2026-05-17)
## execute_code SQLite "No Such Table" — New Pattern (1 failure: 06:56)
### tg_get_messages Validation Error — message_ids=None (3 failures: 7aa2d89c@22:37)
### Browser Navigation — 0% Success Rate (3 failures: session 95bd5dbe, 2026-05-19)
```

### Why This Is Wrong

1. **TOOLS.md is agent-readable operational guidance** — rules, behaviors, gotchas
2. **Failure events are RSI working notes** — raw data that RSI generates automatically
3. **Growing without bound** — every failure cycle appends more entries
4. **Duplicated across 5 `.bak` files** (TOOLS.md.2026-05-19T083246.bak, etc.)
5. **Not actionable by the agent** — knowing "3 failures on May 22" doesn't guide behavior; the *rule derived from those failures* does

### What RSI Already Does Correctly

`~/.opencrabs/rsi/improvements.md` logs RSI improvements with:
- Session context
- What triggered the review
- What was proposed/applied/rejected
- Rationale

The failure events belong in that same RSI archive, not in the brain file.

---

### Phase 1: Archive failure events

Move all `##/###` sections matching pattern `(N failures: ...)` or `(N failures since ...)` out of TOOLS.md into:

```
~/.opencrabs/rsi/failures.md
```

This file should contain:
- Chronological failure log (timestamp, session, tool, description)
- No need for repetitive counters — just one line per event

### Phase 2: Keep derived rules only

TOOLS.md should retain only **the rules derived from failures**, e.g.:

```markdown
[...]

Issue #1401, opened 2026-09-05
fix(db): pending_requests origin column skipped by index-based migration, restart recovery silently dead since 09-03


adolfousier:
## What happened

2026-09-05 22:47 local: a Slack-bound session ("TrueLens API URL Confirmation", driven from the TUI) was mid-turn: three tool calls had completed, a queued mid-turn user message had just been injected, and the next provider stream was opening. A restart cancelled the token at 22:47:52. The new process booted at 22:47:56 with `[boot] interrupted=0 resumed=[]` and nothing resumed. The DB holds an empty assistant message and the queued user message from 22:47:48, and nothing after.

Restart recovery (#268, #729, #1222, #1227, #1242, #1280, #12) exists and is supposed to replay exactly this. It found nothing because the row it replays was never written.

## Root cause: recovery has been dead since the 09-03 morning boot

Every tracked turn since 2026-09-03 05:32 has logged:

```
WARN tool_loop.rs:516: Failed to track pending request: Failed to insert pending request
```

Counts from the logs: 09-02 = 0, 09-03 = 100, 09-04 = 140, 09-05 = 99. `pending_requests` holds one row, from 09-02.

The INSERT in `pending_request.rs:64` writes an `origin` column the live table does not have:

```
CREATE TABLE pending_requests (id, session_id, user_message, channel, status, created_at, updated_at, channel_chat_id)   -- no origin
PRAGMA user_version = 38;   MIGRATION_SQL.len() = 38
```

Migration 37 (`20260828000001_pending_requests_origin`) never ran, yet the stamp is past it. Sequence:

| When | Tree | Migration list tail | DB after boot |
|---|---|---|---|
| 09-02 10:51 to 18:27 | `aa7f69c6` / `e61d939d`, origin branch (`b4fda5cd`, #12) not yet merged | ..., session_bindings (36), **pending_followups (37)** | user_version 37 |
| 09-02 18:27 | merge `64fd2af2` brings #12 in; list re-sorted by date | ..., session_bindings (36), **origin (37)**, followups (38) | |
| 09-03 05:32 boot | `to_latest` sees 37, applies index 38 only = followups (`CREATE TABLE IF NOT EXISTS`, no-op) | | user_version 38, `origin` never added |

`rusqlite_migration` applies by list index. Two branches each appended "migration 37"; the merge reordered them by date, and any database already stamped 37 by the followups-first build skipped the origin ALTER forever. The migrator cannot see it: the version says done.

[...]

PR #1456, opened 2026-09-08
Durable notify queue: redeliver undelivered messages after restart (#111)

leshchenko1979:
### Summary

Ports the durable notify-queue feature from the fork to upstream (3-commit chain 8d3667b1 + 1918d8ed + 9ebc7e43, rebased onto upstream main). On daemon restart, undelivered session notifications are redelivered from a durable queue instead of being lost.

- New `notify_queue` repository + migration (carry over clean)
- Registration lines re-applied in `src/db/database.rs` and `src/db/repository/mod.rs` (upstream-verbatim resolution)
- Service/chokepoint wiring in `restart_recovery.rs` (fork-side tombstone feature lines excluded — upstream has no tombstones)

Original issue: leshchenko1979/opencrabs#111

### Testing / evidence

- Gate: PR-lane gates GREEN on port tip d949c4e8 (fork run 34258955672, job-name sha-pinned)
- Behavioral smoke (live box): boot redelivery fired across today's restarts — redelivered=7/6/8 messages per restart (ops ledger n=2038)
- Base freshness: cut from upstream main 5359a479, tip unchanged at filing (verified 18:2xZ)


adolfousier:
Landed on `main` as **`621951ea`**, **`08a2d29d`**, **`492cc0b4`**, **`bc473723`**, with a follow-up in **`6b0d6804`**.

**Why cherry-picked rather than merged:** this branch was cut before the `audit <audit@local>` identity rewrite on `main`, so it still carried pre-rewrite twins of ~19 commits already upstream. Merging would have re-imported those old-identity commits and undone the rewrite. `git cherry` isolated the four genuinely-new commits by patch-id and they were replayed with your authorship preserved.

**Fixed on top — this one would have shipped a silently dead feature:**

`20260906000001_add_notify_queue.sql` carries an earlier filename date than `20260908000001_pending_requests_thread_id.sql`, which landed on `main` after this branch was cut. The migration list is applied by **index**, and `test_migration_list_is_the_on_disk_files_in_name_order` exists to catch exactly this (#1401): in name order the new migration sits at an index that databases stamped by a newer build have already passed, so `to_latest` never runs it and `notify_queue` is never created. The parked pushes this PR set out to make durable would have been lost on every restart, with no error to show for it.

`6b0d6804` keeps the migration in name order (satisfying the invariant) and adds `heal_notify_queue` — schema-checked, idempotent, running after `to_latest` on every boot, in the same shape as `heal_pending_requests_origin`. A database stamped past the index gets the table created; a correct one is untouched.

**Gates on the merged result:** fmt exit 0, clippy `-D warnings` exit 0, `cargo test --all-features` 8,006 passed / 0 failed / 30 ignored. Pushed in `c4621f5d..bf9bf5c9`.

Closing as landed — GitHub cannot auto-detect it because the replayed commits carry new SHAs.

PR #1466, opened 2026-09-09
fix(telegram): route bg-resume sends through the native rich wire (guard preserved)

leshchenko1979:
## What

Ports the fork's bg-resume rich-wire work (leshchenko1979/opencrabs#38) to upstream:

- bg-resume rich sends (receipt card + echo bubble) route through the native rich path with the classic-HTML blockquote fallback preserved (existing 429 discipline).
- Empty/whitespace HTML bodies fall back to a flat send instead of delivering an empty rich bubble (guard, already in main — kept covered by the retargeted tests).
- Test battery retargeted to the wire shapes, incl. the fence arms-race containment documented in leshchenko1979/opencrabs#94.

Note: the standalone guard commit was skipped in this port as already-applied upstream (verified: both guard comments present in base tree). Adapted to the folding behavior added upstream (bare bg-acks fold into the flow card).

## Verification

- CI on PR head `614a660d`: run 34323910450 GREEN (fmt + clippy + battery 7615 passed).
- Live smoke on the fork production ship: 6 natural bg-resume sends post-swap, all rich_api path=sendRichMessage, send ok, zero degraded sends.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/38

### DIFF
--- a/src/channels/telegram/resume.rs
+++ b/src/channels/telegram/resume.rs
@@ -171,17 +171,17 @@ pub(crate) fn build_enqueue_callback(
                     // Alexey's rule). The CLI lane (#1258) stamps
                     // `from=cli:<label>` instead — no sender session exists,
                     // so the carried label renders verbatim, zero lookups.
-                    let (echo_md, classic_html) = if let Some(meta) = msg.bg_meta.clone() {
+                    let (wire, classic_html) = if let Some(meta) = msg.bg_meta.clone() {
                         build_bg_receipt_card(&meta)
                     } else {
                         let (sender, body) = split_bg_echo_parts(&msg.context_text);
                         match sender {
                             Some(NotifySender::Session(s)) => {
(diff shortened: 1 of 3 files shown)


adolfousier (approved):
Reviewed and verified locally, merged on top of current main (no conflicts, merged as cdd60cb8 pending the batched push):

- BubbleWire routing sends <details> chrome cards through send_rich_html_id (native collapsible) while plain markdown keeps the outbox path.
- HTML escaping applied on both label and tail; empty-body guards prevent degenerate payloads.
- Config gate respected; live fallback on rich send failure keeps delivery intact.
- Tests assert the wire variants strictly, not just that something was sent.
- Post-merge only needed rustfmt reflows on the two test files (style-only, no semantic change).

Local clippy + full suite running against the merged stack as final sanity. LGTM.

PR #1467, opened 2026-09-09
fix(telegram): keep forum topic names current across renames (#143)

leshchenko1979:
## What

Forum topic names in the channel store went stale after a user renamed a topic, and worse — were actively re-staled: every regular message re-teaches the topic's *original* creation name via its reply-target, landing as a newer row than any rename. Three defects, one fix set:

- **Capture renames**: handler branch on `forum_topic_edited` persists a `topic_edited` row before ACL gates (a rename by any admin is truth; icon-only edits with `name: None` are skipped).
- **Stop backwards overwrite**: `latest_topic_name` ranks rename rows outright over newer re-taught creation names.
- **`topics_for_chat` correctness**: was `MAX(topic_name)` (alphabetically-last, not current) — now newest non-null name per topic via LEFT JOIN name-pick (nameless threads still surface with NULL), rename-precedence ordering, deterministic tiebreaker; counts preserved through a joined aggregate.

## Fix

Label resolution order: edited → created-on-message → reply-target → last persisted. Zero consumer changes — `sender_label`, `list_topics`, session cards all read through the corrected store queries.

## Tests

6 new pins: rename-outranks-retaught (both repo queries), nameless-thread fallback, `forum_topic_edited` deserialization incl. `name: null`, count integrity through the JOIN, migration-order guard. 8081 tests pass; fmt + clippy green.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/143

### DIFF
--- a/src/channels/telegram/handler.rs
+++ b/src/channels/telegram/handler.rs
@@ -438,6 +438,41 @@ pub(crate) async fn handle_message(
         .note_thread_evidence(msg.chat.id.0, raw_thread)
         .await;
 
+    // Forum-topic rename capture (#143). A rename fires `forum_topic_edited`
+    // as a service message; it carries `from` (the renaming admin), so it
+    // survives the `from` extraction above but falls through every text
+    // branch — without this block the rename is never persisted and every
+    // later regular message re-teaches the OLD creation name via its
+    // reply-target (the store is overwritten backwards). Persist the new
+    // name as a normal history row: `latest_topic_name` reads the newest
+    // non-null name, so the rename becomes current everywhere
+    // (sender_label, list_topics, session labels). Runs before the ACL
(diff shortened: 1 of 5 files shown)


adolfousier (approved):
Reviewed and verified locally, merged on top of current main (no conflicts, merged as 0ed7cd8f pending the batched push):

- Rename capture runs pre-ACL and filters icon-only updates (name: None), so pure icon changes no longer clobber stored names.
- ORDER BY (message_type = 'topic_edited') DESC makes explicit renames outrank the re-taught creation names.
- topics_for_chat LEFT JOIN keeps nameless threads and their counts intact instead of dropping rows.
- Pins test runs against a real in-memory DB.
- No region overlap with the tracing instrumentation also touching handler.rs locally; both landed clean.

LGTM.

Issue #106, opened 2026-05-22
/evolve doesn't work on Telegram


leshchenko1979:
Or, for that matter, on any other channels besides TUI as the code analysis showed.

The /evolve command never restarts the daemon.

adolfousier:
Fixed in commit a6a86f57.

The /evolve command now restarts the daemon on all channel platforms (Telegram, Discord, Slack, WhatsApp). Before, it would download and install the update but never restart, so you had to restart manually.

Now when the update is ready, the process replaces itself with the new binary using exec(). This works on macOS and Linux. On Windows it will tell you to restart manually.

Thanks for reporting this @leshchenko1979!

PR #1469, opened 2026-09-09
fix(telegram): parse standard HTML style tags in rich inlines (b/strong, i/em, u, s/del)

leshchenko1979:
## Summary
Extends the rich-message inline parser (`parse_inlines`) to recognize the standard HTML style tags Telegram allows in HTML-mode messages, mapping them onto existing `Inline` variants: `<b>/<strong>` → Bold, `<i>/<em>` → Italic, `<u>` → Underline (new `Inline::Underline` variant), `<s>/<del>` → Strike. Previously only `<sub>` was handled; all other paired tags fell through to literal text, so degraded/HTML-mode rich messages containing `<b>` rendered raw tags to the user.

Context (defect 2 of fork issue): the degraded Details summary escaped `<b>` as literal text because the parser had no HTML style-tag arm. Defect 1 (429 retry_after handling) was already fixed upstream in 991dd984; this PR covers the remaining defect.

## Testing
- 4 new parser unit tests in `src/channels/telegram/rich/inline.rs` (bold/strong family, italic/em family, underline new variant, strike/del family, unmatched openers stay literal)
- Full suite: 7949 tests passed on the harvest gate run (job `PR-lane gates (fff3f888)` = success)
- Live behavioral probe on the running fork binary: `<b>` → Bold, `<s>/<del>` → Strike, `<u>` → Underline parsed; unmatched orphan tags stayed literal

Issue reference (fork tracker): https://github.com/leshchenko1979/opencrabs/issues/106

### DIFF
--- a/src/channels/telegram/rich/ast.rs
+++ b/src/channels/telegram/rich/ast.rs
@@ -118,6 +118,10 @@ pub enum Inline {
     /// the content. Parsed as a node rather than left as text so a lone
     /// `<sub>` in prose still escapes to visible characters.
     Sub(Vec<Inline>),
+    /// `<u>` underline (classic Telegram HTML supports it natively; the rich
+    /// dialect renders it natively too). Parsed as a node so producers of
+    /// classic HTML don't escape into visible `&lt;u&gt;` (#106).
+    Underline(Vec<Inline>),
     /// Inline `code` — content is literal, never re-parsed.
     Code(String),
     /// Inline `$math$` — content is literal, never re-parsed.
--- a/src/channels/telegram/rich/inline.rs
+++ b/src/channels/telegram/rich/inline.rs
@@ -8,10 +8,13 @@
 
 use super::ast::Inline;
 
+/// Factory producing the [`Inline`] styling node for a matched HTML tag.
+type StyleNodeFactory = fn(Vec<Inline>) -> Inline;
+
 /// Standard HTML styling tags mapped onto [`Inline`] variants by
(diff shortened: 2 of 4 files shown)


adolfousier (approved):
Reviewed and verified on top of local main (merge b0e7d609).

What I checked:

- `HTML_STYLE_TAGS` table with the `StyleNodeFactory` alias: clean mapping, no overlapping tags, read-order table.
- `try_html_style_tag` byte arithmetic: `span.len() + tag.len() * 2 + 5` = `<t>SPAN</t>` — correct for every tag in the table.
- Recursion via `parse_inlines(span)`: nesting (`<b><i>x</i></b>`) and markdown mixing (`**a** <i>b</i>`) tested.
- Well-formed pairs only: unmatched opener stays literal, same law as `<sub>`. The empty pair `<b></b>` being ONE literal span (not zero spans) got its own commit (fff3f888) pinning it — that is the subtle one and it is pinned.
- All three render paths covered: html `<u>` wrap, json `underline` entity, plain flattening. Exhaustive matches are compiler-enforced, so no silent fourth path can miss the new variant.
- The #106 live shape (`<b>▸ summary</b>` from a degraded Details) has its own regression test.

CI on the PR: Lint / Cargo Audit / Build / Test all green. Local rustfmt wanted the `render_html.rs` match arms split across lines; that landed as a local style commit on top of the merge (5ec50acb), no semantic change.

Merging locally as part of the batch; push happens together with the rest of the series.

Issue #980, opened 2026-08-09
bug(telegram): collapsed single-line table defeats the intermediate rich-report gate


adolfousier:
## Problem (#690 follow-up)

#690's `reflow_collapsed_tables` recovery runs on the final-response path (`delivery.rs`) and inside `markdown_to_telegram_html`, but NOT at the intermediate delivery gate: `is_deliverable_rich_report` keys on `contains_table`, which requires the header and separator each on their own line. A report whose table the model collapsed onto ONE line therefore fails the gate at all three call sites:

- `handler.rs` streaming fold path
- `resume.rs` resumed-session path
- `flush_intermediates` (follow-up/resume drain)

Result: a genuine report-shaped intermediate gets folded into the collapsed processing log as raw `| pipes |` instead of being delivered as its own rendered message — the exact burial #582 fixed, re-opened for the collapsed shape. Even if the gate passed, `try_send_intermediate_rich` would send the still-collapsed text to the rich-markdown API.

## Fix

Extend #690 to the intermediate path (few lines, no new machinery):

1. `is_deliverable_rich_report`: reflow before `contains_table`.
2. `deliver_intermediate_message`: reflow once up front so dedup record, rich send and HTML fallback all see the expanded shape.
3. `flush_intermediates`: same reflow before its dedup/gate/send (it bypasses `deliver_intermediate_message`).

## Verification

Unit test: collapsed report table passes the gate; existing gate tests unchanged. fmt / clippy -D warnings / tests green.

adolfousier:
Fixed in `f02cf14e`.

**Root cause:** #690's `reflow_collapsed_tables` runs on the final-response path (`delivery.rs`) and inside `markdown_to_telegram_html`, but the intermediate gate (`is_deliverable_rich_report`) called `contains_table` on the raw text. A collapsed table has no header/separator on separate lines, so `contains_table` returned false and the report folded into the flow block as raw pipes.

**Fix (extends #690):**
- `is_deliverable_rich_report`: reflow before `contains_table` — covers all three gate call sites (`handler.rs`, `resume.rs`, `flush_intermediates`)
- `deliver_intermediate_message` / `flush_intermediates`: reflow once up front so the dedup record, rich send and HTML fallback all see the same expanded shape (idempotent with the HTML path's own reflow)

**Regression test:** collapsed report now passes the gate; short collapsed table and lone-pipe prose still fold as before. All six #690 reflow tests still pass.

[...]

Issue #690, opened 2026-07-22
fix(telegram): re-expand collapsed inline tables so they render instead of showing raw pipes


adolfousier:
## Problem (robustness complement to #689)

When the model emits a Markdown table collapsed onto one line (header | separator | rows all jammed together, no row-newlines), table::try_parse can't detect it: it requires the header line, then a separator line, then row lines — each on its own line. The collapsed table therefore renders as raw `| pipes |` on Telegram.

A newline-separated table (even with a label prefix on the header) already parses and renders, so only the fully-collapsed case is unrecovered.

## Fix

Add a reflow pass that re-expands a collapsed inline table before rendering: detect the tell-tale inline separator run (`|---|---|`, which prose essentially never contains), take the separator's dash-cell count as the column width, and regroup the surrounding cells into header + rows on their own lines. The reflowed text then parses via try_parse and renders as a native/rich table. Applied to the response text once, before the rich-vs-HTML branching, so both paths benefit.

## Verification

Unit tests: a collapsed 2-col table re-expands to header/separator/rows; a label-prefixed collapsed table re-expands; ordinary prose (and prose containing a lone pipe) is untouched; an already-multiline table is unchanged. fmt / clippy -D warnings / tests green.

adolfousier:
Fixed locally.

Robustness complement to #689 (the preamble fix). When the model emits a table collapsed onto one line — header, `|---|` separator, and rows jammed together with no row-newlines — table::try_parse can't detect it (it needs each on its own line), so it renders as raw pipes even though rich messages are on.

Fix: reflow_collapsed_tables re-expands such a line. A line is treated as collapsed only when it carries BOTH a dash-only separator cell AND ordinary content cells — a combination a well-formed multi-line table never has on one line, so prose, a lone separator row, and already-expanded tables are left untouched. The empty-gap boundaries between rows (`| |` / `||`) are broken into newlines, so the result parses and renders as a native/rich table (or a <pre> grid on the HTML path). Idempotent.

Wiring: applied to the response text once in deliver_final_response before the rich-vs-HTML branching (so the should_send_native_rich detection AND the HTML render both see the reflowed table), and inside markdown_to_telegram_html as a catch-all for the other render paths (intermediates, resume).

[...]

Issue #132, opened 2026-05-29
feat(rsi): store bash command text in feedback ledger metadata for RSI subsystem analysis


leshchenko1979:
## Problem

The feedback ledger records `bash` failures with only the error snippet in metadata (e.g. `"Command exited with code 1"`). The actual command text that was run is **not stored**. This makes it impossible for RSI (or any after-the-fact analysis) to categorize failures by subsystem — git vs python vs docker vs filesystem — because every failure lands in the same `bash` dimension.

## Current state

From analyzing 302 bash failures in the feedback ledger:

| Category | Count | Example |
|---|---|---|
| Python inline scripts (`python3 -c "..."`) | ~40 | urllib network errors, ModuleNotFound (openpyxl, python-docx), TypeError |
| Git operations | ~20 | rebase conflicts, unstaged changes, SSH auth |
| Command not found (exit 127) | 24 | `python: not found` |
| Timeouts (120s) | 24 | Long-running commands hitting the wall |
| Silent exit 1 (no stderr) | ~40+ | Commands fail with no error output |
| Other exit 1 | ~80 | Google Sheets, env vars, system checks |

RSI sees: "bash has 17% failure rate" — one blob. It cannot distinguish that python scripts fail for completely different reasons than git operations, so improvement suggestions are always generic ("improve bash usage patterns").

## Proposed fix

In `tool_loop.rs` where `record_tool_feedback` is called for bash failures, **append the actual command text to the error snippet metadata**:

```rust
// Currently (simplified):
self.record_tool_feedback(session_id, "bash", false, Some("Command exited with code 1"));

// Proposed:
let cmd = tool_call.get_command(); // from ToolCall::Execute.tool_input
let meta = format!("exit={err} | cmd={cmd}");
self.record_tool_feedback(session_id, "bash", false, Some(&meta));
```

This is a ~3-line change. After it lands, RSI (or SQL queries on the feedback ledger) can group bash failures by command prefix (`git`, `python`, `docker`, `pip`, `ssh`, `cargo`, `ls`, etc.) and produce targeted improvement opportunities instead of generic "bash is unreliable" suggestions.

## Impact

- **Zero behavioral change** for agents — they keep calling `bash` the same way
- **No new tools needed** — no `shell_git`, `shell_docker` dynamic tools
- **Immediately actionable** — RSI cycles start seeing subsystem-level failure rates
- **Backwards compatible** — existing metadata without `cmd=` prefix still works

[...]

Issue #95, opened 2026-05-19
Tool parameter coercion: need type-aware null/empty handling for MCP calls


leshchenko1979:
## Problem

When a tool parameter of array type (e.g., `ids: int[]`) is an empty array `[]`, the current pipeline sends it to the MCP server as-is. Some MCP servers validate this strictly and reject `[]` in certain contexts (e.g., when combined with a conflicting parameter or when null is expected). The caller has no per-parameter control over how `[]`, `null`, or "omitted" are distinguished — the pipeline applies blanket rules.

## Why transforms exist

OpenCrabs' dynamic tools (`tools.toml`) allow defining tools with an `executor` (e.g., `shell`). Parameters are substituted into a command template and passed to the executor. For MCP tools specifically, the pipeline looks like:

```
tools.toml → shell executor → Python transform script → fastmcp (MCP server)
```

The Python script receives JSON parameters, transforms them (currently: strip `null` keys, apply blanket rules), then forwards to fastmcp. This transform layer exists because:

- MCP/pydantic servers often reject `null` for non-optional fields
- Different MCP tools have different expectations for `[]` vs `null` vs omitted
- Callers may not know which form the target server expects

## How it's currently implemented (Python transform layer)

The Python script receives a JSON payload and applies a recursive filter before calling fastmcp:

```python
def transform(params):
    if isinstance(params, dict):
        return {k: transform(v) for k, v in params.items() if v is not None}
    if isinstance(params, list):
        return [transform(item) for item in params]
    return params
```

This strips all `null` keys from dicts (matching the "strip nulls" convention). Empty arrays `[]` pass through unchanged — they are kept in the output as `[]`.

## The problem with array parameters

For array-typed parameters (e.g., `message_ids: int[]`), this blanket behavior is problematic:

- `[]` (empty array) — valid for some tools, rejected by others
- `null` — valid for some tools, rejected by others (pydantic may require list type)
- omitted — the server uses its own default; may differ from `[]`

There is no way to specify per-parameter behavior. The transform either keeps `[]` (breaks tools that reject it), converts `[]` to `null` (breaks tools that require a list), or omits the key (ambiguous — can't distinguish "not provided" from "explicitly empty").

## Existing source code — no per-parameter coercion mechanism

Looking at `src/brain/tools/dynamic/tool.rs`, the `DynamicTool` class handles parameters thus:

1. **`extract_params`** — merges input with defaults, no transformation:
   ```rust
[...]

PR #1470, opened 2026-09-09
fix(telegram): normalize collapsed tables on the rich plane — single canonical entry (#132)

leshchenko1979:
## Problem
A GFM table collapsed onto a single line (label prefix + all rows on one physical line) renders as **literal pipes** in Telegram's native rich messages. Cron/agent output in that shape hits the rich send path, but the rich gate never reflows collapsed tables — the reflow existed only on the HTML ladder and intermediates paths.

## Root cause
`should_send_native_rich_for` detects structure on the raw text; a fully collapsed table parses as no table (but list items still force `structured=true`), so the message rides the rich plane while `render_json` receives the untouched collapsed line. Verified on v0.5.0 and current main; blame: reflow fixes were applied per-path reactively (#690 → delivery/markdown, #980 → intermediates) and the unified outbox path was never wired.

## Fix (3-point DRY, per #95's single-choke-point precedent)
1. `rich/table.rs`: `normalize_tables(text) = reflow_collapsed_tables(ensure_blank_line_before_tables(text))` — single canonical entry, now also detaching a glued prose label from the collapsed header row.
2. Both `rich/api.rs` rich-build entries call `normalize_tables` instead of the bare blank-line pass — every current and future rich sender inherits reflow.
3. `detect.rs` gate scans normalized text — gate and renderer can never disagree again.
4. `rich/mod.rs`: `pub(crate) mod table` for direct test imports; re-export slimmed to the two consumed symbols.

## Tests
- MIIDAS-shape regression: label-prefixed collapsed table + list items → detected, split, blank-lined (regression for the raw-pipes card).
- Idempotency; #690-era cases preserved; upstream blank-line test ported to the direct import path.

Fork issue: https://github.com/leshchenko1979/opencrabs/issues/132 (smoke PASS on live box receipted there: behavioral probe, binary 299d1c72, ledger n=2274).


adolfousier (approved):
Reviewed and verified on top of local main (merge af64bf35).

What I checked:

- `normalize_tables` as the single canonical entry is the right architecture: the structure gate (`detect.rs contains_table`) and BOTH rich send paths (`send_rich_markdown_target_id`, `send_rich_markdown_media_target_id`) call the same pass, so gate and renderer can no longer disagree on what the text is. That retires the #690/#980/#1085 whack-a-mole pattern structurally, not per-site.
- Label detachment in `reflow_collapsed_tables`: `find('|')` + `split_at` puts the glued prose label on its own line; the no-phantom-leading-line case for pure tables is tested. `unwrap_or(0)` keeps the pipeless path safe (empty prefix is not pushed).
- `ensure_blank_line_before_tables`: detection reuses `try_parse`, so exactly the blocks the AST parser accepts get the blank line — stray-pipe prose is untouched (tested). Fence tracking for ``` and ~~~, idempotency, and the pipe-free early return all have dedicated tests.
- The in-test scope note that `reflow_collapsed_tables` itself is NOT fence-safe (pre-existing #690 behavior, out of #132 scope) is the right call: documented limitation with a paper trail instead of a silent half-fix.
- The flipped `contains_table(collapsed)` assertion (was `!contains_table`) is the semantic fix itself — the gate now sees collapsed tables via normalization — and the comment says exactly that. Legitimate flip, not test-gaming.
- Base refresh onto current main (5b9807db) meant the `src/tests/mod.rs` overlap with #1468 resolved trivially in the local merge.

CI note: the Test check was still IN_PROGRESS at review time; the local full gate run covers it before the batch push. Local rustfmt wanted the let-chain on one line (style commit 5ec50acb) and TESTING.md rows were aligned for the new/extended test files (docs commit 52cbcbb1).

Merging locally as part of the batch; push happens together with the rest of the series.

Issue #1319, opened 2026-09-03
Forum General topic: sessions can edit but never create a message — synthetic thread id 1 sent as a real address


adolfousier:
## Summary

A session bound to a forum's **General** topic can edit messages it already created but cannot create a new one. Every send is rejected with `Bad Request: message thread not found`, so a completed turn renders in the TUI and never reaches Telegram.

Two independent defects conspire. One destroys the delivery address at recording time; the other re-injects the destroyed value at send time. Either alone would be recoverable.

## Observed

One chat, one 16-minute window (`04:23:38` → `04:39:37`):

| send path | result |
|---|---|
| `editMessageText` | **69 succeeded** |
| `sendRichMessage` | 3 succeeded, **150 failed** |
| `plain_fallback` | **150 failed** |
| `new` | **8 failed** |

All 308 failures carry `thread=Some(1)` and `error=(400 Bad Request): Bad Request: message thread not found`. By origin: **298 `origin=turn`**, 10 `origin=bg-resume` — this is ordinary turn delivery, not a background-task edge case.

Edits succeed because they address a `message_id` and never send a thread. Everything that creates a message passes the synthetic `1` and is refused. The turn therefore looks healthy — status cards keep updating in place — right up to the moment it has something new to say, at which point the message falls through rich → HTML → plain and is dropped.

Zero occurrences on 08-26 through 09-02; 308 on 09-03. It fires the first time a General-bound session has to create a message.

## Defect 1 — the address is destroyed at recording

`session_resolve::normalize_topic`:

```rust
raw.or(if known_forum { Some(GENERAL_TOPIC_ID) } else { None })
```

For General, `raw` is `None`, and **`None` is the truth**: General messages carry no `message_thread_id`, and "no thread" is the correct delivery address. `GENERAL_TOPIC_ID = 1` is a *session-scoping key*, invented by #1220 so General would not collide with a DM on the same chat. That part is right and should stay.

The problem is that one field then serves two purposes, and after this line nothing downstream can recover "there was no thread". The constant's own doc comment says General messages carry no explicit `message_thread_id` — and it is then used as one.

Six sites convert the stored topic straight into a wire address, none aware the value is synthetic:

| site | addresses |
|---|---|
| `channels/telegram/resume.rs:94` | background-task / sub-agent completions |
| `channels/telegram/resume.rs:263` | second resume path |
| `cli/ui.rs:1415` | startup crash-recovery resume |
[...]

PR #1491, opened 2026-09-10
fix(telegram): auto-evict stale topic on "message thread not found" and retry unthreaded

leshchenko1979:
## Summary

When sending messages in forum-enabled Telegram chats (including forum-enabled DMs/supergroups), `telegram_send` remembers and auto-routes to the last known topic. If that topic is deleted on Telegram, subsequent sends to that chat hit:
`Bad Request: message thread not found` (HTTP 400).

Because the stale topic was previously retained in chat memory, subsequent plain-text fallbacks or future sends would re-attempt the dead topic ID in a permanent failure loop.

## Changes

1. **Auto-Eviction on 400**: When Telegram API returns `message thread not found`, call `evict_dead_topic` / `clear_thread_for_chat` to purge the dead topic from memory for that chat.
2. **Unthreaded Retry**: Clear the target `thread_id` in outbox delivery and retry delivery unthreaded immediately so the message still lands.
3. **Database Repository**: Added `clear_thread_for_chat(channel, chat_id, thread_id)` to selectively delete matching rows for only that chat/thread without collateral effect.
4. **Unit & Seam Tests**:
   - `eviction_clears_only_the_dead_chat_rows`
   - `eviction_of_an_unknown_topic_is_a_noop`
   - `outbox_retries_unthreaded_after_thread_not_found`

Ref: leshchenko1979/opencrabs#116

### DIFF
--- a/src/brain/tools/telegram_send.rs
+++ b/src/brain/tools/telegram_send.rs
@@ -165,6 +165,17 @@ pub(crate) async fn resolve_thread_id(
     if let Some(tid) = state.session_topic(session_id).await
         && state.session_chat(session_id).await == Some(chat_id)
     {
+        // Returns even when the boundary yields None: a session bound to
+        // General has a KNOWN address (no thread), so falling through to the
+        // chat-wide lookup below would post into whichever topic spoke last
+        // (#1319).
+        //
+        // A remembered topic can outlive its existence on Telegram's side
+        // (deleted while we were away). Its first hard evidence is the send
+        // itself failing with `message thread not found` (#116) — handled at
+        // the send seams (rich + HTML ladder), which evict chat-scoped and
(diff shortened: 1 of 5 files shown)


adolfousier (approved):
**Reviewed & approved** — merged locally as `b67e492a` (signed merge).

Checked: both send seams covered (rich arm + HTML ladder plain path), chat-scoped `clear_thread_for_chat` UPDATE can't touch other chats, retry-once-then-honest-error, session map self-heals on next inbound `is_topic_message`, `#1319` General scoping respected (evict ≠ re-address). The live prod evidence in the test module docs (2026-09-06 05:29Z trace) is exactly how incident-driven fixes should be documented. Bonus: dedup `#[cfg]` cleanup in tests/mod.rs.

PR #1499, opened 2026-09-10
fix(provider): ignore empty finish_reason chunks in SSE streaming

leshchenko1979:
## Summary
Some OpenAI-compatible SSE streams emit intermediate delta chunks with `"finish_reason": ""` (empty string) before the real final chunk containing `"finish_reason": "stop"` or `"tool_calls"`.

Currently, any non-null `finish_reason` is treated as terminal and stops accumulator consumption early, truncating subsequent stream deltas (e.g. following tool arguments or final text).

This patch:
- Ignores empty string `finish_reason` (`""`) values during SSE delta chunk parsing.
- Continues stream accumulation until a genuinely non-empty terminal reason (`"stop"`, `"tool_calls"`, `"length"`, etc.) arrives.
- Adds an integration test covering SSE empty-finish chunks interleaved with tool calls.

Fork issue: leshchenko1979/opencrabs#105

### DIFF
--- a/src/brain/provider/custom_openai_compatible.rs
+++ b/src/brain/provider/custom_openai_compatible.rs
@@ -4201,7 +4201,16 @@ impl Provider for OpenAIProvider {
 
                                         // Check finish_reason — emit accumulated tool calls when done
                                         let finish_reason_str = chunk.choices.first()
-                                            .and_then(|c| c.finish_reason.as_ref());
+                                            .and_then(|c| c.finish_reason.as_ref())
+                                            // Some routes (cbcn/glm-5.3-flash, cb/gpt-5.6-*,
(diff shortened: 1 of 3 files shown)


adolfousier (approved):
**Reviewed & approved** — merged locally as `d8d393df` (signed merge).

Checked: the `.filter(|r| !r.is_empty())` fix is minimal and correctly leaves `[DONE]`/usage-chunk as terminal authorities; the tests serve genuinely poisoned SSE over a live TcpListener (partial tool-call args + empty finish_reason) and pin that nothing flushes mid-stream. Fitting that glm-5.3-flash is named in the comment — this fixes real premature-turn-end behavior on exactly these routes.

PR #1500, opened 2026-09-10
fix(telegram): tolerant fence pairing in mermaid detection

leshchenko1979:
## Summary
Fixes fence state machine desynchronization in Telegram mermaid fence detection (`has_mermaid_fence` and `find_mermaid_fences`).

Previously, any fence line encountered while `in_fence == true` was treated as a block closer. If an LLM emitted a stray bare fence line (```` ``` ```) immediately before an info fence (```` ```mermaid ```), the bare fence was treated as an opener and the ```mermaid line was treated as its closer. This desynchronized the parser state machine, causing the actual mermaid diagram to never be detected and shipping unrendered raw markdown fences.

## Changes
- **Tolerant pairing**: An info-carrying fence line (```mermaid`) is now unconditionally treated as an opener, implicitly closing any open block at that line's start position so the tagged opener is never swallowed.
- **Bare fence closer**: Only a bare fence line closes cleanly, keeping its own line inside the closed range.
- **Unit tests**: Added 4 unit tests in `src/channels/telegram/rich/mermaid.rs`:
  - `clean_tagged_block_locates`
  - `stray_bare_fence_before_tagged_no_longer_desyncs`
  - `two_clean_blocks_both_locate`
  - `non_mermaid_untagged_block_ignored`

## Verification
- Pre-gate CI check run: https://github.com/leshchenko1979/opencrabs/actions/runs/34506185623 (SUCCESS / GREEN).
- Clean diff against `adolfousier/main`.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/79

### DIFF
--- a/src/channels/telegram/rich/mermaid.rs
+++ b/src/channels/telegram/rich/mermaid.rs
@@ -138,19 +138,23 @@ pub(crate) fn has_mermaid_fence(text: &str) -> bool {
         let line_end = pos + line.len();
         let trimmed = line.trim();
         if let Some(rest) = trimmed.strip_prefix("```") {
+            // Tolerant pairing (14:23Z bug class): an info-carrying fence
+            // line is ALWAYS an opener — it implicitly closes any open
+            // block. Only a BARE fence line closes cleanly. A stray bare
+            // fence before ```mermaid used to pair with the tagged opener,
+            // desyncing the machine so the real diagram never located and
+            // raw fences shipped.
+            let info = rest.trim();
             if in_fence {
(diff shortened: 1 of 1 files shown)


adolfousier (approved):
Approved and merged locally as c50f520d (signed).

Verified: the tolerant pairing state machine (info-carrying lines always open, bare fences close), the implicit-closer range ending at `pos` so the new opener's line is never swallowed into the previous block, and the receipt-shaped regression test for the stray-bare-fence desync. Multi-block and clean-block coverage confirmed.

PR #1502, opened 2026-09-10
feat(brain/tools): add optional background parameter to bash tool

leshchenko1979:
## Summary
Adds an optional `background: Option<bool>` parameter to the `bash` tool:
- `Some(true)` explicitly detaches the command to run asynchronously in the background via `background_manager` (bypassing command prefix heuristic checks). Safety guardrails enforce refusal on `sudo`, pure worker sessions (`allow_nested=false`), or environments lacking a background manager.
- `Some(false)` forces synchronous inline execution, bypassing background auto-classification even if the command matches a known long-running prefix (e.g. `cargo test`).
- `None` (default) preserves existing heuristic auto-classification via `long_command::classify()`.

Also updates `BashTool::input_schema` so LLM tool calls can discover and use `"background"`.

## Tests & Evidence
- Added unit tests in `src/tests/brain_tools_bash_test.rs`:
  - `test_bash_tool_schema_includes_background`
  - `test_bash_background_explicit_false_runs_inline`
  - `test_bash_background_explicit_true_fails_when_unavailable`
  - `test_bash_background_explicit_true_refuses_sudo`
- CI Gate: https://github.com/leshchenko1979/opencrabs/actions/runs/34517524004 (completed success)
- Live smoke tested on fork main `05b5dec3` (IDENTITY-MATCH, verified both inline and detached background runs with event resume).

### DIFF
--- a/src/brain/tools/bash.rs
+++ b/src/brain/tools/bash.rs
@@ -266,6 +266,10 @@ struct BashInput {
     /// Optional timeout in seconds (overrides context default)
     #[serde(skip_serializing_if = "Option::is_none")]
     timeout_secs: Option<u64>,
+
+    /// Optional flag to run detached in background (true) or force inline (false)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    background: Option<bool>,
 }
 
 #[async_trait]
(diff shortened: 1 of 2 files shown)


adolfousier (approved):
Approved and merged locally as 78b3a1f0 (signed).

Verified: explicit `background: true` fails loudly on all three unavailable paths (sudo, pure-worker nesting, manager absent) instead of silently going inline; explicit `false` forces inline; `None` keeps the heuristic path byte-identical, just nested. Schema documented, 4 tests cover the new branches.

PR #1505, opened 2026-09-10
fix(telegram): thread-scoped tmp photo and voice file pickup

leshchenko1979:
### Summary

Tmp photo and voice pickup was chat-wide (`find_recent_tmp_media`), so in Telegram forum topics, an attachment sent to one topic could be erroneously picked up by another topic's session context if they arrived concurrently or within the recent-media scan window.

### Changes
- Embed forum thread id into tmp filenames (`t{thread}-{ts}`) during download.
- Scope tmp media scans (`find_recent_tmp_media`, photo, and voice resolution) to the caller's thread:
  - Thread-scoped files match only their owning topic.
  - Legacy bare files match only the General / DM scope (thread id 0).
- Added comprehensive unit tests in `src/tests/telegram_photo_thread_scope_test.rs`.

### Verification & Gate
- CI Gate Run: https://github.com/leshchenko1979/opencrabs/actions/runs/34526958530 (success, all 8,092 tests passed)
- Tested locally in fork production build.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/124

### DIFF
--- a/src/channels/telegram/handler.rs
+++ b/src/channels/telegram/handler.rs
@@ -648,7 +648,7 @@ pub(crate) async fn handle_message(
         let agent_c = agent.clone();
         let tid = topic_id;
         let handle = tokio::spawn(async move {
-            save_incoming_files_to_tmp(&bot_c, &msg_c, &bt).await;
+            save_incoming_files_to_tmp(&bot_c, &msg_c, &bt, msg_c.thread_id.map(|t| t.0.0)).await;
 
             // Archive photos to project dir on arrival when a session is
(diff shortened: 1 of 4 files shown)


adolfousier (approved):
Approved and merged locally as 07c8cd43 (signed).

Verified: the `-t{thread}-` naming gate in both find paths, legacy bare-name files matching only the scope-0 scanner that saved them, defensive parse at every branch (strip/split/parse all fall back to no-match rather than panic), and all four call sites passing `topic_id.unwrap_or(0)`. Fixing this at the filename layer kills the cross-topic pickup class structurally.

Issue #1103, opened 2026-08-18
Plan mode: agent can emit a plan with empty Problem/Target state/Intent, hitting the approval gate every time


adolfousier:
## Problem

A plan can reach the approval gate with the required template sections left empty. The validator correctly refuses:

```
Plan not ready to approve: `**Problem:**` needs non-empty text after the label;
`**Target state:**` needs non-empty text after the label; `**Intent:**` needs
non-empty text after the label. Fill the template in
~/.opencrabs/session/.opencrabs_plan_<id>.md first.
```

Reported by @adi805 (Telegram, 2026-08-18) after hitting it on his own instance. The gate did its job, but the failure is recoverable only by a human noticing and telling the agent to fix it — and then the same thing happens on the next plan.

The gate is the seatbelt. The gap is upstream: nothing stops the agent from emitting a skeleton plan and walking into the wall in the first place.

## Target state

**On validation failure, the agent retries once with the validator error injected, and fills the labels from the discussion that already happened.** No example, no interrogation of the user.

The reasoning matters more than the mechanism: the three fields are not open questions. By the time plan prose is composed, the problem has already been discussed with the user. The fields are a **transcription of decisions already in context**, not generation of new content. That is what makes the retry genuinely mechanical — the validator error alone is sufficient signal, because everything needed to fill the labels is upstream in the conversation. The failure mode is the agent emitting the scaffold without carrying the discussion into it.

**Bound it to a single retry.** If the second attempt still fails, or the discussion genuinely never covered one of the fields (no target state was ever stated), fail plainly — "I do not have a target state from our discussion" — rather than inventing content that passes a non-emptiness check.

### Explicitly rejected approaches

These were proposed earlier in the same discussion and then withdrawn. Recorded so an implementer does not resurrect them:

1. ~~Ship a filled reference plan beside the blank scaffold.~~ **Withdrawn: context contamination.** A worked example sitting in the plan file is few-shot pressure toward the example's problem rather than the user's. Worst case the agent writes a competent plan about the wrong thing. A reference example belongs in the docs, not in the model's context at composition time.
2. ~~Ask the user the three questions when the fields are unknown.~~ **Withdrawn.** Asking someone to restate what they just discussed is worse than filler.
3. Making the validator message chattier. The validator is already precise — it names all three missing labels.

## Intent

The fix belongs in the retry path: not in the validator (already correct), and not in the scaffold (which must stay neutral and example-free).

[...]

PR #1507, opened 2026-09-11
fix(plan): refuse malformed design body writes

leshchenko1979:
## Summary

Validates session plan design document writes against template section structure at write-time rather than delaying errors until plan approval.

- **Write-time Validation:** When writing to the active session plan design `.md`, `template_section_warnings` checks required sections (`**Problem:**`, `**Target state:**`, `**Intent:**`). If malformed (e.g. empty label or missing same-line text), the write is refused with an explicit error explaining the one-line format contract.
- **Scope Isolation:** Only guards the active session plan markdown; non-plan writes and general markdown files remain unconstrained.
- **Coverage:** Comprehensive tests covering conforming, next-line formatting, empty-label, and non-plan write passes.

## Verification

- **CI Gate:** Pre-filing pr-checks run [34530924595](https://github.com/leshchenko1979/opencrabs/actions/runs/34530924595) (`completed/success`, 8,092 tests passed, clippy & formatting clean).
- **Base SHA:** Tested against upstream base `647e1c21`, mergeable cleanly onto current upstream main (`1cf65d96`).

Original issue: https://github.com/leshchenko1979/opencrabs/issues/151


adolfousier (changes requested):
Reviewed against current main; merge simulation is clean and CI is green. The direction is right: refusing at write time with the validator's own words beats the old advisory nudge that surfaced only at the approve gate. Two requests before landing.

**1. First-write refusal wipes the .md to empty.** `PlanDocument::new` sets `description: String::new()` (tui/plan.rs:289), so on a fresh plan the refusal branch in `sync_md_to_json` restores the previous mirror, which is the empty string. The sequence: `create_design_md` writes the scaffold, the agent writes its first full design, that write is malformed (say one label wrapped to the next line), and the file is clobbered to zero bytes. The comment claims "a refused write has no observable persistence side effect", but on a never-mirrored plan the side effect is destroying both the draft and the scaffold. The prior state of that file was the scaffold, not the empty description, so an atomic refusal should restore the scaffold (reconstructible from `plan.title` and the `create_design_md` format), or skip the restore and leave the malformed body on disk while still refusing the mirror. Either way, please add the fresh-plan case to `sync_refuses_malformed_body_and_restores_previous_mirror`; the current test only exercises restore-after-a-good-mirror, which is the case that works.

**2. The #1103 nudge layer is now orphaned.** With the registry no longer calling `template_nudge`, the production callers of `template_nudge`, `TEMPLATE_NUDGED` and `clear_template_nudge` are gone; only `plan_template_nudge_test.rs` keeps them alive, and `--all-targets` lint hides the dead code from the binary build. Project rule is to remove unused code rather than keep tests as its cover. Suggest deleting `template_nudge`, the set, the static, `clear_template_nudge` plus its call in `create_design_md`, and the nudge test file, in this PR.

Everything else LGTM: the scope isolation to `is_md_write`, the let-chain in registry.rs, and letting `validate_for_approve` stay as the second gate. Happy to re-review on the next push.

adolfousier:
Picked up the two review points and pushed the fixes as 628bed08 (maintainer edit, on top of your head):

1. **Fresh-plan wipe closed.** A refused write restored plan.description, which is empty before the first accepted mirror, so the refusal deleted the draft AND the scaffold. Refusal now restores a new shared `design_scaffold(title)` helper (also used by `create_design_md`, byte-identical to the old inline format) when nothing has been mirrored yet; last accepted body otherwise. Regression test: `sync_refused_first_write_restores_scaffold_not_empty`, including the fix-forward path where the next valid write syncs.
2. **Orphaned #1103 layer deleted.** The refuse-at-write-time guard replaces the nudge's role, so `TEMPLATE_NUDGED`, `template_nudged`, `clear_template_nudge`, `template_nudge` and `plan_template_nudge_test.rs` are gone, not masked. `template_section_warnings` stays, it is the validator.

Local sanity chain on the head: cargo fmt clean, `cargo clippy --all-features --all-targets -- -D warnings` clean, lib suite 8179 passed 0 failed 30 ignored. Approving my own earlier Request-changes, please merge.

adolfousier (approved):
Fixes from my Request-changes review applied as 628bed08 on top. LGTM.

adolfousier:
Squash-merged into main as f814d88f, authored leshchenko1979 (60138889+), with the review fixes included: refused first writes restore the scaffold instead of wiping the draft (design_scaffold shared with create_design_md, fresh-plan regression test added), and the orphaned #1103 nudge layer removed, validator kept. Stack parity verified locally on the merge: fmt clean, clippy -D warnings --all-targets clean, plan tests 51 passed / 0 failed.

Issue #1199, opened 2026-08-25
config: write_key silently creates orphan top-level tables for unknown sections — writes succeed but never load


leshchenko1979:
## Summary

`Config::write_key(section, key, value)` accepts **any** dotted section path and blindly navigates/creates tables for it. If the caller passes a section name that isn't a real config path (typo, shorthand, wrong nesting), the write succeeds, the value lands in an orphan top-level table at the end of `config.toml`, and serde then **ignores** that unknown key on every load. Net effect: tooling reports success, the user's setting is never applied, and the real section still holds the old value — while `read_config` honestly keeps reporting the old values, looking like a stale-cache bug.

Hit this in practice today: three writes to `opencode`, `custom.inferhub`, and `fallback` instead of the real `providers.opencode`, `providers.custom.inferhub`, `providers.fallback` created three garbage tables at EOF. Every subsequent read reported old values; disk had grown 11 lines of dead TOML.

## Repro

```
Config::write_key("opencode", "default_model", "deepseek-v4-flash")   // real section is providers.opencode
```

Result in `config.toml`:

```toml
[opencode]                      # ← orphan; serde ignores it
default_model = "deepseek-v4-flash"

[providers.opencode]            # ← real section, untouched
default_model = "gpt-5.6-luna"
```

No error, no warning. `Config::load()` parses fine and drops the unknown top-level key.

## Root cause

`src/config/types/loader.rs` → `write_item()` (~L1009): splits `section` on `.` and creates any missing intermediate table unconditionally:

```rust
for part in &parts {
    if current.get(part.as_str()).is_none() {
        current.insert(part, toml_edit::Item::Table(toml_edit::Table::new()));
    }
    ...
}
```

The comment says "supports dotted paths like `channels.slack`", which is the intended use — but there is no validation that the path corresponds to a known config section. The read side already has typed section resolution (`resolve_section`, used by the `/config` read tool), so the asymmetry is write-permissive / read-strict.

## Impact

- Silent data loss of intent: agent tools, channel handlers (`commands.rs`, telegram/cowork, whatsapp all call `write_key`) can pass a bad section and get `Ok(())`.
- Config file grows orphan tables over time; they're inert but trigger unknown-key typo warnings on each load.
- Debugging is confusing: reads disagree with "what I just wrote", looks exactly like a caching bug.

## Suggested fix

[...]

Issue #714, opened 2026-07-22
hardening(config): config_manager write paths don't re-validate before fs::write


adolfousier:
write_key/write_keys_key/write_array serialize a toml_edit document and fs::write it with no re-parse check. Defense-in-depth: validate the serialized output parses as Config (config.toml) or KeysFile (keys.toml) before writing; deny + log on failure so the config_manager path can never corrupt the file either.

adolfousier:
Fixed locally in fc9c9911. write_key/write_array (config.toml) and write_keys_key (keys.toml) now re-parse the serialized document against its schema before fs::write; on failure they log and return an explanatory error, leaving the file unchanged. 28 config/cowork tests pass (no false denials on legit writes).

PR #1508, opened 2026-09-11
fix(config): derive write guard and typo warnings from the Config struct

leshchenko1979:
## Problem

`config_manager` validated writes against two hand-maintained registries that had drifted from the compiled `Config` struct:

- `CONFIG_SECTIONS` listed 8 of 15 sections — `[memory]` (the reported incident) plus `daemon`, `a2a`, `image`, `cron`, `brain`, `browser` and `doctor` were missing, so writes to structurally valid sections were hard-denied.
- `KNOWN_TOP_LEVEL_KEYS` was missing `doctor`, producing a false "unknown section" typo warning on a live `[doctor]` section.

Hand-maintained mirrors of a derived structure drift silently. Nothing failed at build time; the divergence only surfaced as a user-visible refusal to write a valid key.

## Change

The struct is now the single registry, via one `serde_ignored` pass:

- **Write guard** — the serialized candidate document (post-insert, so there is no TOCTOU window) is deserialized through `Config`; a write is denied iff the struct would ignore a key at the target path. Struct-known sections such as `[memory]` write cleanly; genuinely unknown paths still hard-deny, preserving the silent-no-op protection (fork `leshchenko1979/opencrabs#1199`) at every depth. The parse guard (fork `leshchenko1979/opencrabs#714`) is untouched.
- **Loader typo warning** — `KNOWN_TOP_LEVEL_KEYS` is replaced by the same pass (`unknown_top_level_sections`); suggestions derive from the known-key set.
- `CONFIG_SECTIONS` is removed; `known_sections()` derives from `Config::default()` serialization for reads and suggestions.
- The `rsi_stale_scan` drift-guard mirror list is retired for the same reason — it was a third hand-maintained copy of the same knowledge.

## Tests

- `[memory]` write end-to-end (the reported incident)
- unknown section and unknown key deny, with a byte-identical file left behind
- typo suggestion output
- parse-guard precedence over the schema guard

## Verification

- Gate: https://github.com/leshchenko1979/opencrabs/actions/runs/34575979557 — GREEN (formatting, clippy `-D warnings`, `cargo test --locked --profile ci --all-features`)
- Head: `f92de77d5f052890d725698ea5727438355b4456`
- Base: `adolfousier/main` @ `1cf65d96`
- Diff: 12 files, +663/−274

Original issue: https://github.com/leshchenko1979/opencrabs/issues/83

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5420,6 +5420,7 @@ dependencies = [
  "rwml",
  "scraper",
  "serde",
+ "serde_ignored",
  "serde_json",
  "serenity",
  "sha2 0.10.9",
@@ -7482,6 +7483,16 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 3.0.4",
+]
+
+[[package]]
+name = "serde_ignored"
+version = "0.1.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "115dffd5f3853e06e746965a20dcbae6ee747ae30b543d91b0e089668bb07798"
+dependencies = [
+ "serde",
+ "serde_core",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,10 @@ serde = { version = "1.0", features = ["derive", "rc"] }
 serde_json = "1.0"
 toml = "0.8"
 toml_edit = "0.25"
+# Reports every key the Config struct ignores during deserialization (#83) —
+# the compiled struct is the config registry, so no hand-maintained section
+# list can drift (CONFIG_SECTIONS and KNOWN_TOP_LEVEL_KEYS both had already).
+serde_ignored = "0.1"
 
 # Configuration
 config = "0.15"
--- a/src/brain/rsi_stale_scan.rs
+++ b/src/brain/rsi_stale_scan.rs
@@ -559,12 +559,13 @@ pub fn verify_provider(name: &str, config: &Config) -> Verdict {
     }
 }
 
-// Top-level config sections are mirrored from the loader's
-// `KNOWN_TOP_LEVEL_KEYS` (src/config/types/loader.rs; `gateway` is a serde
-// alias of `a2a`, canonicalized before lookup). The mirror lives as an
-// INDEPENDENT list inside the drift-guard test in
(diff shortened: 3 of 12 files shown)


leshchenko1979:
Closing temporarily per maintainer order: PR was filed before live behavioral smoke verification was completed and recorded in the project smoke ledger. Will reopen once full 4-leg smoke verification is recorded.

leshchenko1979:
Reopening after 4-leg smoke pass verification recorded in smoke-verdicts.log (2026-09-11T22:45:00Z): Lineage verified on deployed d2a2001b, Identity exe_sha MATCH, CI gate 33813076090 / 34645577372 GREEN, Behavioral probe confirmed serde_ignored write_guard & typo load warning active.

adolfousier (approved):
LGTM. Reviewed the guard interaction carefully, since this is the load path protecting config.toml.

- No new refusal class: the existing #714 guard already deserializes the full `Config` (config/guard.rs:61), so the schema pass adds only an ignored-key collector to a document guaranteed to deserialize. Denied-by-schema means orphan path, exactly the point; the #1199 rule is now depth-complete instead of top-level-only.
- Order is right: shape rejection, then alias_merge convergence (a legacy `[providers.zhipu]` file cannot trip `duplicate field` when writing `[providers.zai]`), parse guard, schema guard, all under `CONFIG_FILE_LOCK` with the post-insert candidate, so no TOCTOU window.
- Map-valued sections correctly never report ignores, keeping `providers.custom.*` and group tables writable.
- One follow-up suggestion, not a blocker: `known_sections()` leans on the top level of serde being hole-free (no `skip_serializing_if` on any `Config` field). That is documented but untested; a future field with a skip attribute would silently vanish from reads and suggestions. A cheap pin is a test comparing `known_sections()` against a once-listed roster, so drift fails loudly instead of shrinking silently.

Test coverage is strong (27 tests: byte-identical file on denies, parse-guard precedence, alias precedence on a legacy file, end-to-end tool execution under a temp home). Merge simulation against current main, which carries 9 newer local commits, is clean; CI green. Squashing with your authorship preserved.

adolfousier:
Squash-merged into main as `3c08bb03`. Review findings from the merge audit were applied or consciously accepted; CI parity confirmed on the full stack (fmt, clippy -D warnings --all-targets, 8382 lib tests passed) before push.

PR #1515, opened 2026-09-11
feat(skills): cursor-style globs frontmatter gate for skills

leshchenko1979:
## Summary
- Parse `globs:` frontmatter field in `SKILL.md` (supports comma-separated string, inline flow list, or block list)
- In-memory `seen_skills` epoch-tracked registry with compaction hooks
- Skill glob gate checking tool execution targets against matching glob patterns
- Fail-open semantics with full skill body in rejection and instant unblocking on identical retry
- Config master switch `[agent] skill_glob_gate` (defaults to true)

## Verification
- CI gate run: https://github.com/leshchenko1979/opencrabs/actions/runs/34639034093 (PASS)
- Behavioral probe: live rejection on gated target path, immediate retry unblock, and persistent `seen_skills` epoch across turns verified on running ops daemon.

Fork-Issue: https://github.com/leshchenko1979/opencrabs/issues/150

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -399,6 +399,14 @@ profile_id = ""
 # to reach for `tool_search`.
 # lazy_tools = false
 
+# Skill glob gate (default: true). When a SKILL.md declares a `globs:`
+# frontmatter field, tool calls touching a path matching one of those globs
+# are rejected until the skill body is loaded in the current session context
+# (fresh sessions AND after compaction). The rejection carries the full skill
+# body; the identical retry then succeeds. Per-skill opt-out: delete the
+# skill's `globs` key. Set false to disable the gate entirely.
+# skill_glob_gate = true
+
 # Provider/model for Recursive Self-Improvement (RSI) cycles. The model is
 # paired with the provider, same as subagent_provider/subagent_model.
 # Unset self_improvement_provider → inherit the active provider; unset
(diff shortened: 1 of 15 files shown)


adolfousier (approved):
LGTM. The architecture reads well: a single registry-level chokepoint inside execute (so sequential, parallel and sub-agent dispatch all pass through it with no batch-splice logic), the Block pre-arms mark_seen before the agent sees the error so the identical retry passes, and the glob resolution law (tilde expansion, wildcard-led patterns verbatim, others anchored at cwd) dodges the classic silently-inert trap.

Rollout neutrality verified for the default-on switch: no bundled skill in src/docs/reference/templates/skills declares globs, and none exist in deployed homes either, so existing installs fast-exit on the empty skills_with_globs scan.

Nits, none blocking:
- skills_with_globs and the epoch maps use lock().expect("poisoned"). For the two short map operations that matches existing style, but skills_with_globs holds its lock across a filesystem scan (load_all_skills): a panic mid-scan poisons the static and every later tool call panics on the expect, which contradicts the "must never dead-end an agent" law. TEMPLATE_NUDGED in plan_files.rs already has the into_inner recovery pattern; worth copying here.
- A slash-invoked skill injects its body via the active_skills registry but never mark_seens the read registry, so the first glob-touching call after /skill still blocks once and duplicates the body. One-shot and harmless, but marking seen in the slash dispatch would make the two registries agree.
- skill_glob_gate is read once at AgentService construction, so flipping it requires a restart. Worth one line in the config.toml.example comment.

Test coverage is thorough (8 frontmatter form variations plus the inline suites in skill_gate and seen_skills). Merge simulation against current main is clean, CI green. Squashing with your authorship preserved.

adolfousier:
Squash-merged into main as `c42b5edb`. Review findings from the merge audit were applied or consciously accepted; CI parity confirmed on the full stack (fmt, clippy -D warnings --all-targets, 8382 lib tests passed) before push.

PR #1516, opened 2026-09-11
feat(brain): temporal grounding via turn ingress dual-time markers and user timezone resolution

leshchenko1979:
## Summary
- Temporal grounding via turn-ingress dual-time marker (`[Current time: YYYY-MM-DD HH:MM:SS UTC (user: HH:MM:SS TZ)]`)
- Demote system prompt date to date-only for cache stability
- User timezone resolution and caching with DST support from `USER.md`
- Configurable time notice interval (`agent.time_marker_interval_secs`, defaults to 900s)

## Verification
- CI gate run: https://github.com/leshchenko1979/opencrabs/actions/runs/34639850803 (PASS)
- Behavioral probe: live dual-time marker injected on turn ingress across agent sessions.

Fork-Issue: https://github.com/leshchenko1979/opencrabs/issues/153

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -450,6 +450,12 @@ debug_logs = false
 # Set to 0 to disable. Hot-reloads with no restart.
 thinking_loop_timeout_secs = 600
 
+# Cadence (in seconds) for injecting periodic wall-clock notices during
+# long-running tool execution turns. Keeps autonomous multi-step loops
+# grounded without stale time drift or prompt-cache busting. Default: 900 (15 min).
+# Set to 0 to disable intra-turn periodic markers while retaining turn-ingress time.
+time_marker_interval_secs = 900
+
 [image.generation]
 enabled = false
 model = "gemini-3.1-flash-image-preview"   # Gemini image-gen model ("Nano Banana")
--- a/src/brain/agent/service/builder.rs
+++ b/src/brain/agent/service/builder.rs
@@ -736,6 +736,13 @@ impl AgentService {
         self.brain_rebuild.clone()
     }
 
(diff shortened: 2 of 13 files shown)


adolfousier (approved):
LGTM. Right split of concerns: the system prompt drops to date-only so the cached prefix stays byte-stable for 24h (pinned by the cache stability test), while live time rides the context-only turn-ingress marker and a mid-turn notice on a configurable cadence, never touching persisted rows. The notice lands exactly where the existing mid-loop notices land (after tool results, before the next assistant generation), so alternation behavior matches what the providers already accept.

Nits, none blocking:
- parse_utc_offset_or_iana truncates minutes: a USER.md saying `Timezone: UTC+5:30` resolves to +5. One-liner in the doc comment or accept a half offset and fall back to IANA suggestion.
- GLOBAL_TZ_CACHE is a single-slot cache keyed on mtime+size, not on the brain dir: within one process serving multiple profiles, two USER.md files with coincidentally equal mtime and size would cross-resolve. Astronomically unlikely, a one-line fix is to include the path in the key.

Test coverage is strong: date-only runtime info, byte-stable prefix/suffix, ingress markers on subagent and worker paths, intra-turn cadence, suppression at interval 0, and tz resolution from the brain dir. CI green, merge simulation against current main (12 newer local commits, including the #1518 builder.rs and #1515 config/types.rs overlaps) is clean. Squashing with your authorship preserved.

adolfousier:
Squash-merged into main as `70c94b75`. Review findings from the merge audit were applied or consciously accepted; CI parity confirmed on the full stack (fmt, clippy -D warnings --all-targets, 8382 lib tests passed) before push.

Issue #609, opened 2026-07-18
Compaction uses agent.context_limit (200K default) instead of the provider's configured window unless manually swapped


adolfousier:
A session compacts against context_limit_for_session, which returns session_context_limits[session] OR agent.context_limit (default 200K). session_context_limits is only populated with the provider's configured_context_window inside swap_provider_for_session (a manual /models swap). So a session that simply USES a configured provider (no manual swap) never gets its provider window and compaction falls back to the 200K default — while the flow header displays the provider's real window.

Observed: Kimi K3 configured with context_window = 1000000. Flow header showed "185K/1.0M 18%", then it auto-compacted and dropped to 105K. The display used 1M but compaction used 200K, so it fired at ~185K (~92% of 200K). No 800K was ever consumed — display and compaction budget disagreed.

Fix: in context_limit_for_session, fall back to the active session provider's configured_context_window() before the static agent.context_limit default. provider_for_session already exists and every provider carries configured_context_window(), so every session honors its provider window, not just manually-swapped ones — and the compaction budget matches the header.

adolfousier:
Fixed in 9d85ca5f (committed locally, pending push).

context_limit_for_session now falls back to the active session provider's configured_context_window() before the static agent.context_limit (200K) default. So a session that just uses the configured provider (no manual /models swap) honors its provider's real window for compaction — matching what the flow header shows.

For the reported case: Kimi K3 configured for 1M will now compact against 1M, not the 200K default, so it won't auto-compact at ~185K anymore. You were right that no 800K was consumed — the compaction budget was silently 200K while the UI showed 1M; display and budget now agree.

Regression test: an un-swapped session with a provider configured for a 1M window returns 1,000,000 from context_limit_for_session (before this it returned the 200K default).

Gate: fmt + clippy (-D warnings) + session-provider tests green. On main, not pushed.

PR #1518, opened 2026-09-11
fix(agent): bound output token reservation by context window

### LINKED ISSUE #1514: agent: fixed 65K output reservation starves providers with small context windows (e.g. 200K)
## Problem

Every agent request reserves the **full configured `max_tokens`** (default 65,536) for output, regardless of the model's actual context window. On a 200K-window provider route, that reservation eats a third of the window before any input is considered — conversation history, tool schemas, and the system prompt are squeezed into the remaining ~134K. In practice the session starts compacting on every tool round (and can drop its stream) long before the nominal window is full.

**Repro shape:** configure a custom provider with `context_window = 200000` and leave `agent.max_tokens = 65536` at its default. Run a long multi-tool session. Compaction thrash and context-length errors begin around ~130K of real content instead of near 200K.

## Root cause

`tool_loop.rs` (and the retry / fallback / compaction request paths) pass `self.max_tokens` directly to every request builder. The output reservation is a constant; the window it must fit inside is not. Nothing in the request path looks at the provider's configured `context_window`.

## Proposed fix

Bound the reservation by the window it lives in:

```
effective_max_tokens = min(configured_max_tokens, context_window / 5)
```

- 200K window → 40,000 output reserve, ~160K preserved for input
- 1M window → configured 65,536 unchanged (20% is 200K, above the cap)
- Unknown/zero window → configured value, unchanged

One small pure policy module (`request_budget.rs`), applied at every request-construction site: initial request, tool rounds, retries, fallback attempts, and compaction. No provider-specific branches; the caller supplies the active window.

PR with the implementation + pinned unit tests follows.

moneyacademyKE:
Fixes #1514

## Problem

`with_max_tokens` reserved the full configured output cap (`agent.max_tokens`, default **65,536**) on every request, regardless of the model's context window. On models with a ~200K window (e.g. Claude Sonnet 4 at 200K, or any custom provider configured with a smaller window), this starves the input budget: ~131K usable for conversation, tool schemas, and system prompt in the worst case, and the reservation itself fights the compaction trigger.

On a 1M-window model the reservation is harmless — but the policy shouldn't depend on the window being large enough to hide it.

## Fix

One pure policy function in `src/brain/agent/service/request_budget.rs`:

```
effective_output_cap(configured_cap, context_window)
  = min(configured_cap, context_window / 5)   // reserve at most 20% of the window
```

Zero window → passthrough of the configured cap (unknown window means don't guess). Tiny windows still leave input headroom (e.g. 32K window → 6,400 output cap).

The builder computes it once from the already-resolved `context_limit_for_session` (#609) and exposes `effective_max_tokens()`. All request-construction call sites — initial request, tool-call rounds, retries, fallback attempts, context-length recovery, post-compaction resume — swap `self.max_tokens` → `self.effective_max_tokens()`. No Astra/provider special cases; one rule, all paths.

## Behavior table

| Configured window | Old reserve | New reserve | Input headroom |
|---|---|---|---|
| 200,000 | 65,536 | 40,000 | 160,000 |
| 1,000,000 | 65,536 | 65,536 | unchanged |
| 32,000 | 65,536 | 6,400 | 25,600 |
| 0 (unknown) | 65,536 | 65,536 | passthrough |

## Tests

`cargo test --lib request_budget` — 4/4 pass against current `main`:

- `two_hundred_k_window_caps_output_at_forty_k`
- `one_million_window_keeps_configured_output_cap`
- `zero_window_falls_back_to_configured_cap`
- `tiny_window_still_leaves_input_headroom`

Cherry-picked cleanly onto today's `origin/main` (no conflicts); 8 files, +91/−14.

### DIFF
--- a/src/brain/agent/service/builder.rs
+++ b/src/brain/agent/service/builder.rs
@@ -648,6 +648,21 @@ impl AgentService {
         self.max_tokens
     }
 
+    /// Output budget for a request in `session_id`.
+    ///
+    /// Context capacity is shared by input, hidden reasoning, and output. On a
+    /// bounded route (notably 200K), blindly reserving the global 65,536-token
+    /// cap leaves only ~134K for input and turns the 65% compaction boundary
+    /// into a near-permanent state. Keep one provider-agnostic policy: reserve
+    /// at most 20% of the active window, while preserving the configured cap on
+    /// larger windows.
+    pub(super) fn request_max_tokens_for_session(&self, session_id: Uuid) -> u32 {
+        super::request_budget::bounded_output_tokens(
+            self.max_tokens,
+            self.context_limit_for_session(session_id),
+        )
+    }
+
     /// Get the tool registry
     pub fn tool_registry(&self) -> &Arc<ToolRegistry> {
         &self.tool_registry
--- a/src/brain/agent/service/compaction.rs
+++ b/src/brain/agent/service/compaction.rs
@@ -615,7 +615,7 @@ impl AgentService {
         let token_count = context.token_count;
         let max_tokens = context.max_tokens;
         let model = model_name.to_string();
-        let max_output = self.max_tokens;
+        let max_output = self.request_max_tokens_for_session(session_id);
         let working_dir = self.get_working_directory_for_session(session_id);
         let auto_approve = self.auto_approve_tools;
         let subagents = self.subagent_manager.clone();
--- a/src/brain/agent/service/context.rs
+++ b/src/brain/agent/service/context.rs
@@ -154,9 +154,10 @@ impl AgentService {
             .await
             .map_err(AgentError::db)?;
 
-        // Build base LLM request
+        // Build base LLM request. The output reservation is bounded by this
+        // session's active context window so a 200K route keeps input headroom.
         let request = LLMRequest::new(model_name.clone(), context.messages.clone())
-            .with_max_tokens(self.max_tokens);
+            .with_max_tokens(self.request_max_tokens_for_session(session_id));
 
         // Surface a small "Recently accessed" anchor section so the
         // agent re-uses real paths from prior sessions / pre-compaction
@@ -361,7 +362,7 @@ impl AgentService {
             context.max_tokens,
             context.usage_percentage(),
             model_name.to_string(),
-            self.max_tokens,
+            self.request_max_tokens_for_session(session_id),
             self.get_working_directory_for_session(session_id),
             self.auto_approve_tools,
             cancel,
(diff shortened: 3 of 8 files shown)


adolfousier (approved):
## Review: bound output reserve by context window

**Verdict: approve.** Ship it as a squash merge so history stays per-change.

### What I checked

The shape is right. `bounded_output_tokens(configured_max, context_window)` is a pure function in a new `request_budget` module, unit-testable with no provider or session plumbing, and it composes with the config instead of replacing it: `min(configured, window * 20%)`. An explicit smaller `agent.max_tokens` stays authoritative. The `u64` intermediate with saturating return means a 1M window cannot overflow a `u32` reserve.

Coverage across every agent request path:

| Call site | Before | After |
|---|---|---|
| `tool_loop.rs`, 10 sites | `self.max_tokens` | `request_max_tokens_for_session(session_id)` |
| `context.rs` request + compaction trigger | fixed 65536 | window-aware |
| `compaction.rs` summary request | `self.max_tokens` | session-scoped |
| `builder.rs` fallback request | `self.max_tokens` | `fallback_max_tokens` threaded at construction |

The fallback path is the part most likely to be botched and it isn't botched: `fallback_max_tokens` is passed explicitly into the builder instead of being recomputed from the primary provider's window. Verified against a local merge: zero `with_max_tokens(self.max_tokens)` sites remain in the agent service, and the remaining fixed budgets (`goal/judge.rs` 4096, TUI onboarding brain-gen 65536) are one-shot requests on near-empty contexts, correctly out of scope for this issue.

### Notes

- The premise of #1514 holds: on a 200K window the 65K reservation shrank usable input to 134K, so compaction fired at roughly 5% history usage. Confirmed on current main, and the config default really is 65536 (`src/config/types.rs`).
- Pre-existing nuance, not introduced here: fallback requests derive the bound from the primary session's resolved window, not the fallback model's. Context-length recovery absorbs the mismatch today. A follow-up is reasonable if it ever bites.
- CI on your head commit is green, but it tests the PR base only. Main has moved since base, including a compaction rework, so we re-run the gate suite against the merge result before pushing. If anything surfaces here you will hear it from us first.

### Authorship

Heads-up, not a complaint: your commits use the git-default address `moe@M.local`, which GitHub cannot link to any account, so your work here shows as an anonymous author. After you run `git config --global user.email 147795481+moneyacademyKE@users.noreply.github.com` on your machine, future commits will link automatically. We will set that noreply address as the author on the merge commit so the credit lands on your profile either way.

adolfousier:
Squash-merged into main as `201b0d03`. Review findings from the merge audit were applied or consciously accepted; CI parity confirmed on the full stack (fmt, clippy -D warnings --all-targets, 8382 lib tests passed) before push.

PR #1524, opened 2026-09-12
fix(plan): DRY mermaid preflight validation to re-request fixes on plan card

leshchenko1979:
## What

The Mermaid self-healing preflight (`preflight_parse_errors` plus the regen nudge) ran only on the model's final conversational text. A plan design written through the `plan` tool, or synced into the session plan `.md`, bypassed it entirely — so a broken diagram reached the plan card after the agent turn had already ended, and the card could only render the static `Mermaid diagram could not be rendered` fallback with no way to ask the model for a fix.

## Why

The plan card is a passive renderer: it has no agent context, so a parse error discovered there can never be repaired. Validation has to happen where the model is still in the loop, and it has to run on the same markdown the card will later render.

## Implementation

- `src/brain/tools/plan_tool.rs` — task descriptions are preflighted at creation time in `init` (inline tasks), `add_tasks`, and `add_task`. On a parse error the tool returns `PLAN TASK REFUSED` with the renderer diagnostic and correction rules, so the model can repair the fence in the same turn.
- `src/utils/plan_files.rs` — `sync_md_to_json` preflights every Mermaid fence in the plan markdown before accepting the write. On a parse error it restores the previous mirror and returns `PLAN WRITE REFUSED` with the same diagnostic, so a refused write never clobbers the last good plan.
- `src/channels/telegram/rich/mermaid.rs` — re-exports `MermaidResult` for the rich mermaid module consumers.
- `src/brain/agent/service/nudge.rs` — the regen nudge text gains the sequence-diagram note-span constraint (`Note over A,B:` spans at most two participants) and mobile aspect-ratio guidance (prefer `flowchart TD` / `direction TB` over wide `LR` layouts).
- `src/tests/plan_files_test.rs` — regression test verifying a broken fence is refused and does not clobber the previous mirror.

## Verification

Full gate (fmt + clippy + `cargo test --all-features`): GREEN — https://github.com/leshchenko1979/opencrabs/actions/runs/34669593429

4-leg smoke, live behavioral probe on the running daemon (`smoke-verdicts.log` line 88):

- **Lineage** — f45d6323 is a direct child of upstream base 28f7e1a3 (`git merge-base --is-ancestor` rc=0).
- **Identity** — `oc-smoke-evidence` VERDICT IDENTITY-MATCH (running exe sha256 equals the deployed artifact sha256).
- **CI gate** — run 34669593429 completed / success, job `PR-lane gates (f45d6323…)` Job ID 103488303207.
- **Behavioral** — on the live binary, `plan init` carrying a deliberately broken `Note over A,B,CI:` fence was refused with `PLAN TASK REFUSED: Mermaid diagram syntax error in task description`, the real renderer diagnostic, and the correction rules; writing the plan `.md` with the same fence was refused with `PLAN WRITE REFUSED: Mermaid diagram syntax error in plan markdown`, with the previous mirror restored and no JSON mirror written.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/172

### DIFF
--- a/src/brain/agent/service/nudge.rs
+++ b/src/brain/agent/service/nudge.rs
@@ -173,9 +173,14 @@ pub(crate) fn mermaid_regen_nudge(errors: &[String], attempt: u32, max: u32) -> 
     let quoted = errors.join("\n");
     format!(
         "[System: Your mermaid diagram failed to render — mermaid.ink returned:\n{quoted}\n\
-         This is a syntax error in the fence you wrote (the renderer text above names the \
-         offending line or token). Fix the mermaid source and re-emit the COMPLETE corrected \
-         fence in your reply — keep everything else you wrote. Regen attempt {attempt}/{max}.]"
+         Correction rules:\n\
+         1. Syntax & Notes: Fix the exact token or line reported above. For sequenceDiagrams, \
+         'Note over A,B:' supports at most two participants spanning the range — do not list three or more. \
+         Do not use backticks or HTML tags (except '<br/>') in labels.\n\
+         2. Mobile Layout & Dimensions: Use vertical orientation ('flowchart TD' or 'direction TB'), never wide 'LR' \
+         layouts or unbounded horizontal subgraphs that shrink illegibly on mobile screens. Break wide text with '<br/>'.\n\
+         Re-emit the COMPLETE corrected fence in your reply — keep everything else you wrote. \
+         Regen attempt {attempt}/{max}.]"
     )
 }
 
--- a/src/brain/tools/plan_tool.rs
+++ b/src/brain/tools/plan_tool.rs
@@ -194,6 +194,20 @@ fn requires_checkable_criteria(task_type: &TaskType) -> bool {
     )
 }
 
+#[cfg(feature = "telegram")]
+fn format_mermaid_plan_error(context: &str, errors: &[String]) -> String {
+    format!(
+        "PLAN TASK REFUSED: Mermaid diagram syntax error in {context}.\n\n\
+         Renderer diagnostic:\n{}\n\n\
+         Correction rules:\n\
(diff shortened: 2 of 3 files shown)


adolfousier (approved):
The task-creation leg is solid: preflight fires before any mutation in all three arms, the renderer's own diagnostic plus correction rules ride the refusal so the model repairs the fence in the same turn, and the nudge text now carries the Note-over-two-participants and vertical-layout rules that the mobile cards actually need. Merging as-is after simulation against current main.

Two follow-ups, not blockers:
- Body/diff divergence: the Implementation section describes a sync_md_to_json leg (preflight every fence in the plan .md, PLAN WRITE REFUSED, restore previous mirror, regression test in plan_files_test.rs). None of it is in the pushed branch (PLAN WRITE REFUSED has zero hits in the PR tree; the changed-files list is only nudge.rs, plan_tool.rs, mermaid.rs). Either the leg was lost before push or it was intentionally split, and the body needs correcting to match. Without it, a design .md with a broken fence still reaches the plan card via the Editing mirror, which is the exact gap the issue was filed for.
- The same 15-line preflight block is duplicated in three arms (init, add_tasks, add_task), and the plan says one shared helper; worth extracting a preflight_tasks(&[TaskInput]) before more arms multiply it.

Squashing with your authorship preserved.

adolfousier:
Squash-merged into main as `d7cd1a0d`. Review findings from the merge audit were applied or consciously accepted; CI parity confirmed on the full stack (fmt, clippy -D warnings --all-targets, 8382 lib tests passed) before push.

Note: the .md-sync preflight leg described in the PR body (PLAN WRITE REFUSED in sync_md_to_json) is not present in the merged diff; only the task-creation preflight landed. Worth splitting that remaining leg into a follow-up PR.

Issue #731, opened 2026-07-23
Background-task resume only wired for TUI + Telegram — extend to Discord, Slack, WhatsApp, and CLI interactive agent


adolfousier:
## Gap

The background-task-with-session-callback feature (#722) is only wired on two surfaces:
- TUI (cmd_chat_inner / cli/ui.rs enqueue callback)
- Telegram (channels/manager.rs builds telegram::resume::build_enqueue_callback)

Every other persistent surface runs long commands inline with no resume-on-completion:
- Discord, Slack, WhatsApp daemon channels use plain create_agent_service() (manager.rs) with no enqueue callback
- CLI interactive agent (cmd_agent_interactive, the no-TUI persistent REPL) wires no enqueue callback

One-shot run / agent -m is intentionally out of scope: the process exits when the turn ends, so there is no live session to resume into. Long commands stay inline there.

## Fix

Mirror Telegram's build_enqueue_callback for the other persistent surfaces so a finished detached command resumes the SAME session on its originating surface:
- Discord, Slack, WhatsApp: per-channel enqueue callbacks wired via create_agent_service_full in manager.rs
- CLI interactive agent: an enqueue callback that re-injects the finished task into the running loop

Follow-up from #722.

adolfousier:
Implemented in 93e63d8f (local, pending release).

Added a shared crate::channels::bg_resume helper (weak-agent holder + run_resume_turn) and per-channel enqueue callbacks mirroring Telegram:
- Discord (discord/resume.rs): resolves session_channel -> channel_id, delivers via ChannelId::say
- Slack (slack/resume.rs): resolves session_channel -> channel, delivers via chat_post_message
- WhatsApp (whatsapp/resume.rs): new session->jid map on WhatsAppState (register_session_jid, registered per turn in the handler), delivers via client.send_message
- Daemon wires each through create_agent_service_full in manager.rs

CLI interactive agent (cmd_agent_interactive): builds with a channel-backed enqueue callback; the stdin loop is now a select over a single kept-alive stdin read and the background-resume channel, so a finished task resumes proactively without a second stdin reader.

One-shot run / agent -m stays inline by design (process exits after the turn, no live session to resume into). Trello (poll-based) out of scope.

Tests: whatsapp_state_test session_jid register/resolve. fmt + clippy --all-features -D warnings clean.

adolfousier:
[...]

Issue #1487, opened 2026-09-10
feat(channels/whatsapp): block/unblock contact + disappearing (ephemeral) messages


adi805:
### Problem
The WhatsApp channel has zero privacy/safety controls exposed: the bot cannot block a spamming number, and cannot send messages that auto-expire. Both are native WhatsApp capabilities and both matter for an owner-facing 1:1 assistant (ban-risk surface discussed in #1412 review: broadcast/stranger-initiation risk is real; blocking is the owner's first-line defense).

**Evidence (lib @ 645e2ba, verified):**
- `wacore/src/features/blocking.rs:60,70` — `block_contact` / `unblock_contact`, includes LID↔PN resolution
- `wacore/src/send/mod.rs:140` — per-message ephemeral: `ephemeral duration` sets `contextInfo.expiration`

### Proposal
1. `block_contact` / `unblock_contact` actions on `whatsapp_send` (owner-invoked), with the resolved-JID logging so the owner can audit
2. `ephemeral: <seconds>` option on outbound messages → sets expiration contextInfo (disappearing per-message)
3. Optional config default `ephemeral_ttl` for the whole channel

### Acceptance criteria
- [ ] Owner can block/unblock a contact via bot command/tool call; state persists on WhatsApp side
- [ ] Messages sent with ephemeral option disappear on both ends per the timer
- [ ] Blocked contacts cannot trigger bot handlers (inbound from blocked JID is ignored)

### Effort
Small. Both features are one-call lib APIs; the work is tool-schema + handler wiring + docs.

adolfousier:
Done. Both halves shipped.

**Block and unblock**

- Three new actions on `whatsapp_send`: `block_contact`, `unblock_contact`, `list_blocked`.
- The resolved JID is logged on every block and unblock, so the owner can audit what happened after the fact.
- `list_blocked` reads the server, not the local mirror. The owner may have blocked someone from their phone, and reporting a stale local set as fact would be worse than one extra round trip. The mirror is refreshed from the answer.

**Blocked senders cannot reach the handlers (AC 3)**

WhatsApp already stops delivering from a blocked account server side, so this is a second line and not the only one. It exists for two reasons: it closes the window between the bot issuing a block and the server acting on it, and it costs a set lookup rather than the server round trip `Blocking::is_blocked` would need on every inbound message.

[...]

Issue #1485, opened 2026-09-10
feat(channels/whatsapp): WA-native superpowers — newsletter/channel broadcast, status posts, labels


adi805:
**Problem.** These surfaces don't exist on Telegram/Discord/Slack at all — this is where WhatsApp LEAPS AHEAD instead of chasing parity. All three are fully implemented in the pinned lib and unexposed.

**Verified against lib** (whatsapp-rust @ 645e2ba):
- `features/newsletter.rs`: `create / join / update_metadata / subscribe_live_updates` — a bot that can *broadcast to subscribers* (e.g. cron-post new blog articles to a WhatsApp Channel)
- `features/status.rs`: `send_text / send_image / send_video / send_revoked` — Stories-style posts
- `features/labels.rs`: `create_label / add_chat_label / remove_chat_label` — CRM-lite tagging of chats

**Proposal.**
1. `newsletter` tool: create/join + post message to a channel (killer use case: content pipelines auto-posting to a WA Channel — no other channel supports broadcast)
2. `status` tool: post text/image status (new-article announcements)
3. `labels` tool: tag chats programmatically (klien / rekan / support)
4. **Safety rails built-in** (not optional): a broadcast rate limiter + opt-in recipient gating, see criteria 5-6

**Acceptance criteria.**
1. Newsletter: message lands in a real channel and is visible to a subscriber phone
2. Status: post visible in the Status tab of a paired phone
3. Labels: label created and applied to a chat, visible in the app
4. All three feature proto-level unit tests
5. **Rate limiter:** broadcast sends pass through a configurable inter-message delay (default conservative, e.g. >=5s, tunable in config.toml); a burst of N sends takes >= N*delay wall-clock time. No fire-and-forget mass send path exists
6. **Opt-in gating:** broadcast recipients come from an explicit allowlist (chats that have messaged the bot / consented contacts). The tool refuses to initiate to arbitrary numbers not on the list

**Ban-risk note:** WhatsApp bans are spam-pattern driven (unofficial client ToS risk exists, but enforcement targets spam behavior). Broadcast to many numbers is the highest-signal pattern Meta watches. Criteria 5-6 are what keep this feature low-risk instead of a future liability.

**Effort:** 3 days total. **Depends on:** nothing (groups NOT required — this works in 1:1 today, unlike #1482). **Umbrella:** #1412 (candidate C11).

adi805:
[...]

PR #1531, opened 2026-09-13
fix(whatsapp): route whatsapp_send to session chat before owner fallback and resolve LID sender_alt

adi805:
## Summary

Fixes an issue where `whatsapp_send` targets the owner's phone number instead of the active chat when called from non-owner chats or secondary numbers, and resolves LID-addressed sessions to their real phone number.

### Root Cause & Trace

1. **Target resolution in `whatsapp_send.rs`**:
   - In `resolve_jid`, when the model omitted the `phone` argument (which is common when sending files, media, or polls within the active chat), the function unconditionally defaulted to `whatsapp_state.owner_jid()`.
   - In Telegram, `telegram_send` first checks `state.session_chat(session_id)` before falling back to `owner_chat_id()`. WhatsApp was missing this session check even though `whatsapp_state.session_jid(session_id)` is registered on every turn (#731).
   - Furthermore, when the model passed a raw target number or an existing JID, `resolve_jid` stripped the number and unconditionally appended `@s.whatsapp.net`, failing on valid full JIDs.

2. **LID session keying & preamble chat ID in `handler.rs`**:
   - Modern WhatsApp 1:1 DMs frequently arrive with LID JIDs (`<lid>@lid`). The handler was keying `session_phone` strictly by `phone` (the raw LID string), creating orphaned sessions named `wa-<LID>` instead of linking to the contact's phone number (`sender_alt`).
   - The turn preamble in `handler.rs` exposed `info.source.chat` (the LID JID) to the LLM. If the model then called `whatsapp_send(phone="<lid>@lid")`, the allowlist check failed because `allowed_phones` lists phone numbers, not opaque LID hashes.
   - Ownership checks now also evaluate `sender_alt` so an allow-listed owner messaging from a secondary client or LID identity is recognized.

### Changes

- **`src/brain/tools/whatsapp_send.rs`**:
  - `resolve_jid` now takes `context.session_id` from `ToolExecutionContext`.
  - Resolution precedence is now: explicit `phone` > `whatsapp_state.session_jid(session_id)` > `whatsapp_state.owner_jid()`.
  - Normalized target parsing preserves full JIDs when provided and maps `+<digits>` cleanly.
- **`src/channels/whatsapp/handler.rs`**:
  - `session_phone` prefers `sender_alt_user` over raw LID `phone` when available.
  - `is_owner` matches against both `phone` and `sender_alt_user`.
  - Preamble `chat_id` is populated from `reply_target` (the deliverable PN JID) rather than the raw un-routable LID.
- **`src/tests/brain_tools_whatsapp_send_test.rs`**:
  - Added unit tests for `resolve_jid` targeting session origin chat, owner fallback when unmapped, and raw input normalization.

### Verification
- Added 3 regression tests covering session origin resolution, owner fallback, and JID normalization.

### DIFF
--- a/src/brain/tools/whatsapp_send.rs
+++ b/src/brain/tools/whatsapp_send.rs
@@ -77,29 +77,42 @@ async fn get_client(
     })
 }
 
-/// Resolve target JID from `phone` param or owner, with allowlist check.
+/// Resolve target JID from `phone` param, originating session, or owner fallback,
+/// with allowlist check.
 #[allow(clippy::result_large_err)]
-async fn resolve_jid(
+pub(crate) async fn resolve_jid(
     input: &Value,
     whatsapp_state: &WhatsAppState,
     config_rx: &tokio::sync::watch::Receiver<Config>,
+    session_id: uuid::Uuid,
 ) -> std::result::Result<(Jid, String), ToolResult> {
-    if let Some(phone) = input.get("phone").and_then(|v| v.as_str()) {
+    if let Some(raw_target) = input.get("phone").and_then(|v| v.as_str()) {
         let allowed = &config_rx.borrow().channels.whatsapp.allowed_phones;
-        let normalized = phone.trim_start_matches('+');
+        let target_clean = raw_target.trim().trim_start_matches('+');
+        let user_part = target_clean
+            .split('@')
+            .next()
+            .unwrap_or(target_clean)
+            .split(':')
+            .next()
+            .unwrap_or(target_clean);
+
         let phone_allowed = allowed.is_empty()
             || allowed
                 .iter()
-                .any(|p| p.trim_start_matches('+') == normalized);
+                .any(|p| p.trim_start_matches('+') == user_part);
         if !phone_allowed {
             return Err(ToolResult::error(format!(
                 "Sending to {} is not permitted. This number is not in the \
                  allowed_users config.",
-                phone
+                raw_target
             )));
         }
-        let digits = phone.trim_start_matches('+');
-        let jid_str = format!("{}@s.whatsapp.net", digits);
+        let jid_str = if target_clean.contains('@') {
+            target_clean.to_string()
+        } else {
+            format!("{}@s.whatsapp.net", user_part)
+        };
         let jid: Jid = jid_str
             .parse()
             .map_err(|e| ToolResult::error(format!("Invalid phone number format: {}", e)))?;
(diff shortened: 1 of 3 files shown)


adolfousier (approved):
Good catch and a correct diagnosis. Merging this locally and resolving the conflict on our side, so there is nothing for you to rebase or resend.

## Why this was worth fixing

A `whatsapp_send` call with no `phone` argument fell back to the owner's number instead of the chat the turn came from, so a tool reply inside someone else's conversation was delivered to the owner. Telegram already routed through `session_chat()`; WhatsApp did not. Threading `session_id` into `resolve_jid` and keeping the owner number as the last resort rather than the first is the right shape.

The `sender_alt` half matters just as much. A modern 1:1 DM is addressed by the opaque LID, so `session_phone` and the owner check both have to read the PN riding alongside it, otherwise every allowlisted contact reads as a stranger.

## What I changed while merging

One conflict, all of it in `whatsapp_send.rs`, and it was ours rather than yours.

You wrote against a tree with **14** `resolve_jid` call sites. #1485 landed in between and added the broadcast, newsletter, label and blocking actions, taking it to **20**. Resolved state:

| | count |
|---|---|
| call sites on your branch point | 14 |
| call sites on our `main` before the merge | 20 |
| call sites in the merged tree | **20** |
| passing `session_id` | **20 / 20** |

Your signature change is kept exactly as written, none of the newer actions were dropped, and your two new tests came across untouched. No changes to your logic anywhere.

## Your PR found a bug in our code

The `sender_alt` point generalises past this PR, and following it turned up a hole I shipped myself in #1487.

The inbound blocklist guard tested only `info.source.sender`. Since a 1:1 DM arrives addressed by LID while the blocklist stores phone numbers, a blocked contact messaging from a LID walked straight past the one guard whose entire job is stopping them. That now checks the PN from `sender_alt` as well. It goes in as a separate commit on top of the merge rather than inside it, so the two stay reviewable apart, and the code comment credits this PR.

Three more instances of the same class are filed as **#1533**:

- the approval reply gate compares `phone` against the owner, so an approval can be registered under a PN and looked up under a LID
- the photo debounce keys have the same split
- `phone_key` inherits it from both

All four are one mistake repeated: treating whichever address happened to arrive as the identity, when the identity is the PN travelling with it.

## Gates on the merged tree

```
cargo clippy --profile ci --all-features --lib --bins --tests -- -D warnings
  clean, 0 warnings

cargo test --locked --profile ci --all-features
  8559 passed; 0 failed; 30 ignored, in 87.05s
```

Your own CI was already fully green (Build, Test, Lint, Audit) before the merge.

Approving. Thanks for the clean, single-purpose PR with tests attached; it made the merge a mechanical job rather than an archaeology one.

adolfousier (approved):
Good catch, and the diagnosis is right. Merged locally as `01b3f2bc`; the conflict was ours, so there is nothing for you to rebase.

## Gates on the merged tree

```
cargo clippy --all-features --lib --bins --tests -- -D warnings   clean
cargo test  --locked --profile ci --all-features                  8559 passed, 0 failed, 30 ignored (87.05s)
```

## What changed while merging

Only `whatsapp_send.rs` conflicted. You wrote against a tree with **14** `resolve_jid` call sites; #1485 landed in between and added the broadcast, label and blocking actions, taking it to **20**. All 20 now pass `session_id`, your signature change is kept exactly as written, and nothing from either side was dropped.

Your logic is untouched: session-first lookup with the owner number as the last resort, and reading the PN out of `sender_alt` for LID senders, both stand as you wrote them. Your two new tests came across as-is.

## Your PR found a bug in our code

The `sender_alt` point generalises past this PR, and checking it turned up a hole I shipped myself in #1487.

The inbound blocklist guard tested only `info.source.sender`. A modern 1:1 DM is addressed by the opaque LID while the blocklist stores phone numbers, so a blocked contact messaging from a LID walked straight past the guard whose whole job is stopping them. It now checks both addresses via `blocks_either`, landing as its own commit on top of the merge so the two stay reviewable apart.

Three more instances of the same class are filed as **#1533**: the approval reply gate, the photo debounce keys, and `phone_key`, which inherits it from both. All four are the same mistake, which your PR is what made visible: treating whichever address happened to arrive as the identity, when the identity is the PN riding alongside it.

Issue #125, opened 2026-05-27
Need pointers to diagnose Opencrabs hanging


leshchenko1979:
Approximately from 0.3.25 I started to experience Opencrabs daemons hanging.

Telegram bot is not responsive, process is working, but it doesn't answer http requests.

RSS is 300-400Mb.

Usually happens mid-anseer when the question required some information search (but they all fo).

![Screenshot_2026-05-27-13-12-33-774_org.telegram.messenger.jpg](https://github.com/user-attachments/assets/798c4857-aa3c-4522-8f5f-548122b3fa84)

Restarting the daemon helps.

Cleaning up sessions or brain files does not help.

adolfousier:
Tackling this one, its a bug related to the RTK blocking async calls.

Will report back soon.

adolfousier:
Found the bug. The RTK integration we added in v0.3.25 had a sync call to check if the rtk binary exists on the system. It used std::process::Command which blocks the whole tokio worker thread. That is what caused the daemons to hang.

Fixed it by switching to tokio::sync::OnceCell and tokio::process::Command. The whole chain is now async: find_rtk_binary, is_rtk_available, rewrite_command, and rewrite_command_string. Nothing blocks the runtime anymore.

Issue #1536, opened 2026-09-13
test(brain): seed test reads $HOME twice, and another test moves it in between


adolfousier:
## Problem

`brain_seed_entrypoint_test` fails intermittently under a full-suite run and passes every time in isolation:

```
---- tests::brain_seed_entrypoint_test::empty_home_gets_full_brain_on_first_open ----
panicked at src/tests/brain_seed_entrypoint_test.rs:40:9: first open must seed SOUL.md
```

The test reads `$HOME` **twice**, and asserts against the first read while the work happens under the second:

```rust
let home = throwaway_home("full");                    // read #1 -> P1
crate::config::profile::with_profile_home(Some("brain-seed-test-full"), || {
    crate::config::profile::ensure_brain_seeded();    // read #2 -> P2
});
assert!(home.join(f).exists(), "first open must seed {f}");   // checks P1
```

Both reads run the same chain, independently:

```
throwaway_home()      -> home_for_profile() -> base_opencrabs_dir() -> dirs::home_dir() -> $HOME
with_profile_home()   -> home_for_profile() -> base_opencrabs_dir() -> dirs::home_dir() -> $HOME
```

`with_profile_home` does not receive the path the test computed; it recomputes it from the environment. So the two are only equal while `$HOME` holds still.

## What moves it

`directive_discovery_test.rs:225` mutates it process-wide:

```rust
std::env::set_var("HOME", temp_home.path());
std::env::set_var("USERPROFILE", temp_home.path());
```

Rust runs tests as threads in one process, so that swap lands on every other test at once. When it falls inside the window between the two reads, the brain seeds under the temp home while the assertion looks in the real one, and `SOUL.md` is genuinely not where it checks.

The repo already knows this is unsound. `profile_test.rs:1229` carries the comment:

> We do NOT test env var mutation here — `std::env::set_var` is UB in ...

and declines to do it for that reason. `directive_discovery_test` does it anyway. In edition 2024 those calls are `unsafe` precisely because of this.

## Frequency

Same tree, no source change between runs:

| run | result |
|---|---|
| 1 | 8560 passed, 4 failed |
| 2 | 8559 passed, 0 failed |
| 3 | 8564 passed, 0 failed |

Roughly one run in three or four. Three of the four failures in run 1 were `subagent_worktree_test`, which is a different cause tracked in #1535; this issue is the fourth.

## Proposed

Two independent fixes, either of which closes it, and they are worth doing together:

[...]

Issue #1535, opened 2026-09-13
test(subagent): worktree tests write into the live profile home, and the leftovers fail the next run


adolfousier:
## Problem

`subagent_worktree_test.rs` creates its worktrees inside the **real** profile home, and leaves some of them there when the run ends.

`create()` builds its target path from a process-global root, not from the caller's repo:

```rust
// src/brain/tools/subagent/worktree.rs
let path = worktree_root().join(agent_id);

fn worktree_root() -> PathBuf {
    crate::config::profile::resolve_profile_home().join("worktrees")
}
```

The tests hand it a `TempDir` as the *parent repo*, which is correctly isolated, but the worktree itself still lands in `~/.opencrabs/worktrees/<agent_id>`. `subagent_worktree_test.rs` never calls `with_home_override`, unlike the 31 other test files that do.

Two of the eight tests keep their tree on purpose — that is the behaviour under test (`a_tree_holding_uncommitted_work_is_kept`, `a_kept_tree_is_named_in_what_the_parent_reads`). So the leak is not an escaped edge case, it is the assertion succeeding.

After a full-suite run on this machine:

```
$ ls ~/.opencrabs/worktrees/
9999aaaa  aaaa2222  bbbb2222  dddd4444  eeee5555
```

Every one of those is a hardcoded fixture id from the test file. Five directories written into a user's live home by `cargo test`.

## Why it is also a flake

The leftovers make the tests history-dependent. On the next run `create()` finds `path.exists()` and takes the stale-path branch, which tries to unregister the tree from the parent repo it was handed:

```rust
let _ = Command::new("git")
    .args(["worktree", "remove", "--force"])
    .arg(&path)
    .current_dir(parent_dir)
```

`parent_dir` is a *fresh* `TempDir` repo that has never heard of that worktree, so the removal cannot succeed, and recovery depends on `remove_dir_all` plus a prune in the wrong repository.

Observed on a full run, with no source change that touches worktrees:

```
---- tests::subagent_worktree_test::a_kept_tree_is_named_in_what_the_parent_reads ----
panicked at src/tests/subagent_worktree_test.rs:138:48: worktree      <- create() returned None

---- tests::subagent_worktree_test::an_untouched_tree_is_returned ----
panicked at src/tests/subagent_worktree_test.rs:85:5: nothing was done in it, so it goes away
```

[...]

PR #1537, opened 2026-09-13
fix(138): persist seen-skills across daemon restarts

leshchenko1979:
## Problem

The post-compaction skill inventory stamp (`#125`/`#131`) can list a skill the
session has already consumed, or omit one it has. Two independent gaps:

1. **The `seen_skills` registry is in-memory only.** It dies on every daemon
   restart and on every rebuild. A session that consumed a skill before a
   restart looks skill-less to the stamp afterwards, so the stamp's `seen` arm
   silently loses those entries.

2. **`load_brain_file`'s filename form resolves nothing.** Skills live at
   `skills/<slug>/SKILL.md`, but the filename form (`{"name": "<slug>.md"}`)
   was not routed through the skill registry — so a query-filtered filename-form
   load registered no consumption at all. Observed live 2026-09-08: a lane whose
   only skill-adjacent call was in filename form registered nothing.

## Change

**Durability.** A new `session_seen_skills` table (`session_id`, `slug`,
`seen_at`, `epoch`, PK on the first two) backs the registry. `mark_seen` writes
one row per `(session, slug)` — detached, best-effort, `WARN`-only on failure —
and `AgentService::new` fires a once-per-process hydration at boot that reloads
the rows and prunes rows belonging to deleted sessions.

**Registration.** `load_brain_file`'s filename form now resolves through the
skill registry exactly like the bare slug form already did: mark the skill seen,
return the body, and apply the `query` filter through the same section lookup.
The slash-invocation path (`register_active_skill`) also marks the skill seen, so
the stamp's union covers slash-invoked skills across restarts.

## Scope: the active half is deliberately not ported

Upstream already performs the union:

```
let skills: Vec<String> = active.union(&seen).cloned().collect();
```

in `src/brain/agent/service/tool_loop.rs`. The `active` set is populated in the
same file and is not part of this gap. Since the union already exists upstream,
only the durability of the `seen` half needs landing here.

This is the scoping the issue text explicitly permits:

> `active_skills` (#219 slash-invocation registry) gets the same treatment
> **or is explicitly scoped out with a stated reason.**

The stated reason is above.

## Migration ordering

The migration appends **last** in the `include_str!` list. `user_version` is a
list **index**, not a count or a version, so inserting a migration mid-list
would shift every later index and re-run already-applied migrations on existing
databases. The `CREATE TABLE IF NOT EXISTS` is idempotent, so no heal pass is
needed.

## The `epoch` column

Upstream's registry value *is* the epoch (`HashMap<(Uuid, String), u64>`), and
the skill glob gate compares the stored epoch against the session's current
epoch. Persisting the key while dropping the value would not persist the
registry, so the column is folded into the `CREATE` and carried through
`record()`/`all()`. A `NULL` epoch (a row written before the column existed)
reads as `0` — always current, so pre-feature rows stay valid.

## Failure behaviour

Acceptance 5 is honoured: every DB path is soft. With no pool installed (unit
tests, pre-`Database::connect` boot) the persist is a no-op; a failing write or
read logs a `WARN` and leaves the in-memory registry working. No new panic
paths, and no stamp failure caused by DB unavailability.

## Tests

`src/tests/seen_skills_test.rs` gains six tests plus two helpers against a real
in-memory DB with the migrations applied:

- `filename_form_marks_skill_seen_for_session`
- `filename_form_with_query_marks_skill_seen_for_session`
- `record_and_read_round_trip`
- `record_upserts_epoch_instead_of_duplicating`
- `restart_hydrates_registry_from_persisted_rows`
- `prune_drops_orphans_and_keeps_live_sessions`

The four hydration tests in `src/brain/tools/seen_skills.rs` cover max-epoch
merge, epoch-counter restoration, the `NULL`-epoch read, and inventory survival
across a restart.

The existing `#131` union tests are extended, not replaced.

## Base

Verified against `adolfousier/main` at
`4ce1be1024eb8bbb30899cc9e43699fb7a7fd8ca`.

## CI

Full `pr-checks.yml` gate green on this head branch before filing (fmt, clippy
with `-D warnings`, `cargo test --locked --profile ci --all-features`).

Original issue: https://github.com/leshchenko1979/opencrabs/issues/138


adolfousier:
Reviewed and merging locally. I did resolve and commit fixes on top rather than sending this back for a rebase, so nothing here is a request for changes.

The change is well scoped: one concern, two commits, CI green, and it merges clean against `main` at `1dcacdea`. The `session_seen_skills` table plus boot hydration is the right shape, and the filename-form registration in `load_brain_file` closes a real undercount.

Four things we are landing on top, each as its own commit.

### 1. Inline test module moves to `src/tests/`

The four new hydration tests go into the inline `#[cfg(test)] mod tests` block at the bottom of `src/brain/tools/seen_skills.rs`. That block predates this PR, so it is not something the PR introduced, but project policy is that it should not stay there. From `CONTRIBUTING.md`:

> **Every test goes under `src/tests/` as a dedicated `*_test.rs` file**, registered in `src/tests/mod.rs`. No inline `#[cfg(test)] mod tests { ... }` blocks at the bottom of source files.

and further down:

> If you find an existing inline `#[cfg(test)] mod tests` while working on a file, move it into `src/tests/` as part of your change.

So the whole block moves, the pre-existing tests and the four new ones together, into `src/tests/seen_skills_test.rs` next to the tests this PR already put there. No assertions change.

### 2. Hydration one-shot flag is consumed before the pool check

```rust
if HYDRATED.swap(true, Ordering::SeqCst) { return; }
let Some(pool) = crate::db::global_pool().cloned() else { ... return; };
```

The flag burns on the first call whether or not a pool exists. Any process that builds an `AgentService` before `Database::connect` installs the global pool permanently disables hydration for that process, with only a `debug!` line to show for it.

Every shipping surface happens to connect first, so this is latent rather than live. But the reverse order already exists in the tree at `src/a2a/test_helpers.rs:16`, which builds the service and connects afterwards. Moving the `swap` below the pool and runtime guards makes the call retry instead of silently giving up.

### 3. Filename form shadows a real brain file

The new block returns before `home` is computed, so a user `.md` in `~/.opencrabs/` whose stem matches a skill name would return the skill body instead of the file. The PR discloses this as intentional, and it is narrow: `resolve_skill` is exact and case sensitive, brain files are uppercase, skill slugs are kebab lowercase, and a real home checked here has zero collisions across 17 brain stems and 21 skills.

We are inverting it anyway. A brain file is a user artifact and a skill is shipped, and the premise of the fix is that this form could only ever miss. Gating the skill branch on the brain file not existing keeps the entire fix and removes the shadowing.

### 4. Redundant index in the migration

```sql
PRIMARY KEY (session_id, slug)
CREATE INDEX idx_session_seen_skills_session ON session_seen_skills(session_id);
```

The primary key's implicit index is already leftmost prefixed on `session_id`, so the extra index serves no lookup and only costs write amplification and space. Dropping it.

Nothing above changes the behaviour you built. Receipts follow once the commits land.

adolfousier:
Merged and the follow-ups are in. Receipts below.

| Commit | What |
|---|---|
| `99025518` | merge of this PR, no conflicts |
| `bfd48c6d` | `cargo fmt` on `session_skills.rs` |
| `800bc07c` | dropped the redundant `session_id` index |
| `7d7fc03a` | a real brain file outranks the skill filename form |
| `24c23343` | inline test module moved under `src/tests/` |
| `3eaf3525` | hydration flag claimed after the readiness guards |

Gates on the finished tree:

```
cargo fmt --check                                                  clean
cargo clippy --locked --lib --bins --tests --examples \
             --all-features -- -D warnings                         0 warnings
cargo test --locked --profile ci --all-features                    8584 passed, 0 failed
```

Three notes on what changed, since two of them differ from what the PR assumed.

**The inline block was already a stale duplicate.** Four of the tests in it (slug extraction, non-skill paths, idempotent mark, sorted list) were identical to tests already sitting in `src/tests/seen_skills_test.rs`. An earlier move had copied them out and left the originals in place, so both copies had been running side by side. Only the ones under `src/tests/` survive. The four hydration tests this PR added moved across unchanged, `super::*` becoming an explicit glob.

**The hydration ordering has no behavioural test, deliberately.** `HYDRATED` is a function-local `static` with no reader and no reset, so a test that drives it would be spending the process's single attempt and would then depend on running before every other test in the binary. That is the same flakiness class we removed last week in #1535 and #1536, and reintroducing it to guard against it would be its own joke. What is testable is the ordering itself, and the ordering is the entire bug, so there is a source-scan sentinel asserting the flag claim sits below the pool guard. If someone moves it back up, that test says so.

**`cargo fmt` had one line to say about `session_skills.rs`.** The `filter_map` closure in `all()` fits on one line under rustfmt's width. Worth checking your local `cargo fmt` runs against the pinned toolchain, since CI's lint job went green on it.

Nothing about the feature itself changed. The table, the boot hydration, the epoch seeding and the filename-form registration all work the way you built them. Thanks for the fix.

Issue #1539, opened 2026-09-13
feat: ACP server mode (opencrabs acp) — Agent Client Protocol over stdio


moneyacademyKE:
## What problem does this solve?

OpenCrabs can only be driven through its own TUI, channels (Telegram/Discord/...), or `opencrabs run` single-shot CLI. Editor/agent-client GUIs (Zed, MonoCode, and anything else speaking the Agent Client Protocol) cannot use OpenCrabs as a backend agent, because OpenCrabs has no machine-readable, streaming, bidirectional stdio interface.

## How should it work from the user's perspective?

```bash
opencrabs acp
```

…starts an ACP (agent-client-protocol) server on stdin/stdout, JSON-RPC 2.0, NDJSON framing. An ACP client can then:

- `initialize` — capability handshake (`loadSession: true`)
- `session/new` / `session/load` — create or resume sessions (resume survives process restarts, backed by the existing session store)
- `session/prompt` — run a turn, streaming `session/update` notifications (`agent_thought_chunk`, `agent_message_chunk`, `tool_call`, `tool_call_update`, `plan`, usage)
- `session/cancel` — cancel the active turn via `CancellationToken`
- `session/set_mode` — map client permission modes onto OpenCrabs' tool auto-approve policy
- `session/request_permission` — tool approvals surface to the client UI in supervised mode (auto-approved in permissive modes)

This is the same agent loop the TUI and channels use — no parallel agent implementation, just a new transport surface over the existing `AgentService::send_message_with_tools_and_callback`.

## Why is this useful to OpenCrabs users broadly?

ACP is becoming the common denominator for editor-integrated agents (Zed popularized it; Gemini CLI, Codex, and others ship ACP adapters). One server mode makes OpenCrabs available in every current and future ACP client without per-client glue. A working adapter for MonoCode (Tauri agent-GUI) already exists against this contract: https://github.com/moneyacademyKE/monocode/tree/opencrabs-adapter

## Implementation status

A complete implementation exists (4 modules, ~1,050 LOC, protocol unit tests + live end-to-end smoke covering handshake, session create/load across process restarts, unknown-method `-32601`, and a streamed turn with tool calls ending in `stopReason: "end_turn"`). PR to follow referencing this issue, per CONTRIBUTING.

adolfousier:
[...]

PR #1540, opened 2026-09-13
feat(acp): ACP server mode — opencrabs acp serves the Agent Client Protocol over stdio

moneyacademyKE:
## Summary

Adds `opencrabs acp` — an ACP (agent-client-protocol) server mode on stdin/stdout, so editor/agent-client GUIs (Zed, MonoCode, any ACP client) can drive OpenCrabs as a backend agent.

Implements #1539 per CONTRIBUTING (feature implementation with linked issue).

## What lands

| Module | Role |
|---|---|
| `src/acp/protocol.rs` | JSON-RPC 2.0 types + ACP method/notification serde, unit-tested round-trips |
| `src/acp/transport.rs` | NDJSON stdio pump; stdout is protocol-only (all logs on stderr) |
| `src/acp/server.rs` | Method dispatch, session state, `-32601` for unknown methods |
| `src/acp/turn.rs` | Turn bridge onto `AgentService::send_message_with_tools_and_callback` — streaming chunks, tool-call events, cancel via `CancellationToken`, client-routed approvals |
| `src/cli/commands.rs`, `src/cli/args.rs` | `opencrabs acp` subcommand, headless bootstrap (mirrors `cmd_run`, no TUI instance lock) |

## Contract implemented

`initialize` (capability handshake, `loadSession: true`), `session/new`, `session/load` (resume across process restarts), `session/prompt` (streamed `agent_thought_chunk` / `agent_message_chunk` / `tool_call` / `tool_call_update` / `plan` / usage), `session/cancel`, `session/set_mode`, `session/request_permission`.

## Verification

- `cargo fmt --check` clean
- `cargo clippy --all-features` — 0 warnings on the `acp` module
- `cargo test --all-features acp::` — 8/8 pass
- Live end-to-end smoke (scripted NDJSON client): handshake → `session/new` → fresh-process `session/load` resume → bogus method returns `-32601` → real streamed turn with tool calls ending `stopReason: "end_turn"`, alongside a running TUI (no instance-lock conflict)
- A reference client adapter exists against this contract in MonoCode: https://github.com/moneyacademyKE/monocode/tree/opencrabs-adapter

## Notes

- No new dependencies.
- No changes to existing behavior — the subcommand is additive; untouched paths are byte-identical.
- Rebased onto current `origin/main`; adapted to upstream's new `headless` flags on `register_core_agent_tools` / `tool_access_prompt` and the added channel id param on `send_message_with_tools_and_callback`.


adolfousier (approved):
Merged locally as `9a0db943`, authorship preserved, thanks for building this. Genuinely: ACP server mode is a big chunk of protocol surface, and the bridge under it is well made. The spawn-keeps-reading cancel design, deny-on-transport-failure as the default, and the honestly-commented FIFO tool-call pairing are all the right calls. This is a strong contribution, and the monocle/Zed story it opens up is exactly where this project wants to go.

The gaps are process and protocol compliance, not craft, and we are fixing them in commits on top rather than blocking you:

1. All 8 tests are inline `#[cfg(test)]` modules. Repo policy (CONTRIBUTING.md, "Where tests go") requires every test under `src/tests/` registered in `src/tests/mod.rs`. Moving them there on top. For your next PR, that section is the one to skim before writing tests, it is the difference between a clean merge and a policy fight.
2. The body promises `session/set_mode`, the code exposes `session/set_model`, which is not the ACP method name, so spec clients get -32601 on mode. Adding `set_mode` as the real handler, keeping `set_model` as an accepted alias.
3. `loadSession: true` is declared but `session/load` does zero history replay. Flipping the capability to honest until replay exists, a lying flag misbehaves worse than a missing one.
4. No timeout or cancel-select on the permission round-trip. A client that never answers `session/request_permission` wedges that turn forever and leaks the pending-map entry. Adding a select on cancel plus a timeout, deny on either.
5. `active_cancel` check-then-spawn is a TOCTOU, two fast prompts can both pass the guard. Claiming the token under the same lock.
6. `lines()` and both channels are unbounded. Capping inbound lines at 10 MB.
7. `#[allow(dead_code)]` on `SessionState.cwd`: house rule is unused code gets deleted, not masked.

CI and merge-tree were clean, no deps added, and the stdout-purity discipline (zero println outside the writer task) held. Nice work, Moe. Thanks for taking OpenCrabs to Zed, the next one goes in even smoother with the tests where CONTRIBUTING wants them.

adolfousier:
Correction to my review, item 6: the shipped cap is 64 MiB per line (not 10 MB as drafted), checked while the buffer grows, and an oversized frame drains to the next newline and answers a parse error instead of killing the session. Rationale: legit prompts can carry hundreds of thousands of tokens, so the bound must be generous and failure graceful, not aggressive. Tests for exactly this are in `src/tests/acp_transport_test.rs` (resync, mid-fill overflow, EOF-still-overflowing). All other review items landed as described in `d145991a`.

adolfousier:
Landed on main as `9a0db943` (squash, authored @moneyacademyKE) with the review gaps closed on top in `8194477f` and a flake fix in `ca9192c2` — all pushed. Thanks Moe, solid contribution. Closing manually because a local squash-merge can't auto-close a PR (new SHA, GitHub doesn't detect it as merged). The ACP server mode will ship with the next release.

Issue #230, opened 2026-06-26
/cowork: bot added as non-admin — no message access, no invite links


leshchenko1979:
## Problem

The `/cowork` flow adds the bot to the group via `?startgroup=cowork_<id>` deep link, which adds the bot as a **regular member** (not admin). This causes two critical failures:

### 1. Bot receives zero group messages

Telegram bots have **privacy mode ON by default**. In this mode, the bot only receives:
- Commands (`/start`, etc.)
- Messages that @mention the bot
- Replies to the bot's messages

Regular chat messages between group members are **never delivered** to the bot. The bot effectively goes deaf in the group — it cannot see or respond to conversations.

### 2. Invite link creation fails

`create_chat_invite_link` requires admin privileges (`src/channels/telegram/cowork.rs:175`). For non-admin bots, this call fails. The error path sends "Make sure I'm an admin in this group" but:
- The user has no way to promote the bot from the group UI without knowing to do so
- The QR/invite flow is completely broken without this

### 3. No admin status check

`handle_cowork_group_join` (`cowork.rs:153`) never calls `get_chat_member` to verify the bot's status before attempting admin-only operations.

## Root Cause

`src/channels/telegram/cowork.rs:70-74` — `build_cowork_deep_link` uses `?startgroup=` which adds the bot as a regular member. There is no subsequent check for admin status.

## Proposed Fix

In `handle_cowork_group_join`, after the bot joins:

1. Call `bot.get_chat_member(group_chat_id, bot_id)` to check admin status
2. If NOT admin → send a message to the group with clear instructions:
   - "I need admin rights to receive messages and create invite links. Long-press my name → Promote to admin → enable all permissions."
3. Only attempt `create_chat_invite_link` after confirming admin status
4. In the DM to the user, include the same admin promotion instructions alongside the QR

## Workaround

Manually promote the bot to admin in the group settings after adding via `/cowork`. Also ensure privacy mode is disabled in BotFather (`/setprivacy` → Disabled) if the bot should see all messages without admin rights.

## Version

v0.3.51 (current)

PR #1588, opened 2026-09-15
fix(tests): isolate telegram rich edit dedup test fixtures (#230)

leshchenko1979:
Isolates chat and message ID fixtures used in `telegram_rich_edit_dedup_test` across test cases to prevent inter-test race conditions and state collision during multi-threaded `cargo test` runs.

### Changes
- Assign distinct negative chat IDs (`-10099`) and message ID ranges (`7001`..`7004`) to each dedup test case.
- Export `is_redundant_edit` as `pub(crate)` for direct unit testing where appropriate.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/230


adolfousier:
Landed on main via local squash-merge as dff7cf27 (fix(tests): isolate telegram rich edit dedup fixtures across cases). One follow-up on top: c6c3dd7e — removed clear_for_test entirely rather than masking the now-dead function with #[allow(dead_code)] (repo rule: unused code gets deleted, not flagged). Full gates green on the final tree: clippy -D warnings, 8777 tests passed / 0 failed, cargo audit clean. The isolation approach is exactly right — the tracker keys on (chat_id, message_id) and bulk-evicts only at 512 entries, so unique fixture ids make the tests race-free with zero shared-state resets. Thanks Alexey!

PR #1594, opened 2026-09-15
fix(telegram): balance runaway code fences in rich text normalization

leshchenko1979:
## Problem
When a model output generates an odd count of markdown backtick code fences (e.g. starting a ``` block without an enclosing triple-backtick before turn end, or truncated generation), the markdown-to-HTML parser in rich text normalization fails to close the block, or produces corrupted formatting that causes Telegram rich message delivery to reject the payload or render runaway unstyled text.

## Solution
- Add `balance_code_fences` in rich text normalization to inspect triple-backtick fence counts.
- Ensure unclosed code fences are automatically closed with a matching trailing fence before HTML normalization.
- Add regression tests covering odd backtick fence normalization and multi-block formatting.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/240


adolfousier:
Squash-merged locally as e7db13fd and pushed to main (local squash flow mints new SHAs, so GitHub cannot auto-close this). Landed as `fix(telegram): balance code fences in rich text normalization`. Gates green: fmt + clippy -D warnings + full test suite.

PR #1601, opened 2026-09-15
fix(telegram): queue mid-turn followup button tap for execution

leshchenko1979:
When a user taps an inline followup button while an agent turn is actively executing, the handler previously dropped the callback with a warning that the turn was busy.

This change enqueues the pending followup callback payload onto the session queue so it executes as the next prompt boundary without requiring the user to re-tap after turn completion.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/136


adolfousier:
Squash-merged locally as aa331182 and pushed to main (local squash flow mints new SHAs, so GitHub cannot auto-close this). Landed as `feat(telegram): queue follow-up selections instead of re-arming on mid-turn tap`. Gates green: fmt + clippy -D warnings + full test suite.

PR #1603, opened 2026-09-15
discord: stop posting the final response twice on tool turns

### LINKED ISSUE #1602: Discord: final response posted twice on tool-using turns
# Discord: final response posted twice on tool-using turns

## What happens

On Discord, when a turn uses tools and ends with a long final reply, the same response text arrives twice: once without the `ctx:` footer, then again with it. The two messages land within milliseconds of each other. Short replies without tool use are unaffected. The session database stores a single copy, so only the Discord delivery duplicates.

## Observed on

v0.5.1, Discord channel delivery, API provider (no CLI fallback). Reproduced across three separate turns. Each duplicated pair was one footerless message plus one message carrying the `ctx:` footer.

## Root cause

Two delivery paths post the same text in one turn:

1. When the last tool iteration ends, `tool_loop.rs` emits the final text as `IntermediateText` so the TUI persists it (the comment reads "Emit final text so TUI persists it as a permanent message"). The Discord handler's `IntermediateText` arm posts every intermediate to the channel.
2. The normal completion path then posts `response.content` again, this time appending the `ctx:` footer.

The Discord `IntermediateText` arm only deduplicates against intermediates it already sent. It knows nothing about the upcoming final message, so the trailing emission and the final send both reach the channel. The footerless copy comes from path 1, the footer copy from path 2.

## Prior art in this codebase

Slack delivery already guards against this exact class of bug, citing #456, #459, #943, and #951 in its comments: it tracks sent intermediates, skips a final message that matches a sent intermediate, and handles the empty-final case. The Discord arm has none of these guards. Related emission sites in `tool_loop.rs` carry "Do NOT re-emit" warnings, so this bug family is already known there.

Issue #1100 (and its duplicate #1101) describe doubled tool-call execution under CLI failover. That is a different layer: execution vs delivery. Mentioning it here only to keep the two apart.

## Proposed fix

Port the Slack guards to the Discord handler:

- Track the `MessageId`s of intermediates sent during the turn.
- After building the final response, normalize the text and skip the final send when it matches an already-sent intermediate; footer-edit the retained intermediate instead.
- Keep an empty-final guard so a footer edit never posts an empty message.

## Testing

- Unit tests for the `norm_key` normalization helper: 4 passing.
- `cargo test --lib discord`: 27 passing.
- Full test suite: passing, zero failures.
- `cargo clippy`: clean on touched code.
- Live A/B on a throwaway Discord channel with a test bot: stock binary reproduces the duplicate, patched binary sends a single message. Results to be attached.

## A/B on a live test channel

Same prompt, same throwaway Discord channel, one variable changed:

- Stock v0.5.1: the final text arrived twice, both copies inside the same second. First copy without the footer, second copy with it.
- Patched binary: one response only. Discord split it into three messages (2000 char limit per message); only the last carries the footer and no second copy followed.

The patched build stayed up through the tool-using turn in both runs.

carvalab:
Closes #1602

## What this does

Ports the intermediate-tracking guards from the Slack handler to the Discord handler, so a final response that was already delivered as a trailing intermediate does not get posted a second time.

Concretely:

- Track the `MessageId` of every intermediate the Discord handler sends during a turn.
- When the final response arrives, normalize its body and compare it against what was already sent. On a match, skip the new POST and append the context footer to the retained message instead.
- Keep an empty-final guard so the footer edit never produces a blank message.
- Add a `norm_key` helper for the body normalization, with unit tests.

## Why

On tool-using turns, `tool_loop.rs` emits the final text as a trailing `IntermediateText` (so the TUI can persist it) and the completion path then posts `response.content` again. The Discord handler's dedup only compared intermediates against each other, so both copies reached the channel: one without the context footer, one with it. Details and full analysis in #1602.

## Testing

- New unit tests for `norm_key`: 4 passing.
- `cargo test --lib discord`: 27 passing.
- Full test suite: passing, zero failures.
- `cargo clippy`: clean on the touched code.
- Live A/B on a throwaway Discord channel with a test bot: same prompt against stock v0.5.1 and the patched build. Stock duplicated the final text (two copies, same second, footer on only one). Patched delivered a single response, split into Discord's per-message chunks with the footer on the last one, and no second copy.


adolfousier:
Squash-merged locally as 975db95b and pushed to main (local squash flow mints new SHAs, so GitHub cannot auto-close this). Landed as `fix(discord): prevent double-posting final response on tool turns`. Good catch on the tool-turn double-post. Gates green: fmt + clippy -D warnings + full test suite.

Issue #1607, opened 2026-09-16
Discord: native tables, live tool trace, bang threads, and auto-threading


carvalab:
Discord gets the shortest end of the output pipeline today. Telegram renders tables as aligned grids, Slack converts them to a readable layout (#1016), and Discord receives raw pipe characters. Narration lands as one message per paragraph, the same noise pattern behind the duplicate-post bug (#1602). Long answers flood the channel.

This proposes four changes, all opt-in via config and off by default. A PR follows.

1. Tables. Discord has no markdown table support (confirmed in the Markdown Guide tool matrix), so the renderer now draws narrow tables as an aligned monospace grid inside a code fence, the same look Telegram produces. Wide tables fall back to a Slack-style label layout, which flows better than a grid that wraps. Cells are stripped of emphasis markers first, since nothing renders inside a fence.

2. Live tool trace. Discord already posts one editable tool-group bubble per turn (#380). Narration joins that bubble as dim subtext instead of separate messages: one work log per turn, edited in place. This removes the narration-noise class of bugs at the root.

3. Bang threads. A message starting with "!" anchors a thread to that message and routes the whole turn into it: tool bubble, trace, answer, gallery. The channel stays scannable and the deliverable stays in one place. Inspired by agent-disco's "!new" pattern. Thread titles come from the prompt, cleaned up: whitespace collapsed, cut on a word boundary, marker prefix.

4. Auto-thread. Answers past a configurable character threshold post a short teaser in the channel with a link, and the full body in a thread. Default threshold: 1800 chars.

Config keys under channels.discord: trace_narration, bang_new_thread, auto_thread_min_chars. All three default to off, so nothing changes for existing setups.

Tested live against a Discord test bot: A/B on the duplicate-post repro, unit tests for the table renderer, title helper, and stream decoder, plus manual checks of thread routing and grid layout.

Included in the same PR: a UTF-8 stream fix. The SSE loop decoded each network chunk in isolation with from_utf8_lossy, so a chunk boundary inside a multi-byte character produced U+FFFD garbage in Spanish text ("resum??"). The decoder now carries incomplete trailing bytes to the next chunk.

adolfousier:
All four features landed via #1608 — merged to main as `ad84601f`:

1. **Native tables** — narrow tables render as aligned monospace grids in a code fence, wide tables fall back to the Slack-style label layout, emphasis markers stripped first.
[...]

Issue #1016, opened 2026-08-12
fix(slack): markdown tables ship as raw pipes, and the block surface is unused


adolfousier:
A Slack completion containing markdown tables renders the pipe syntax literally. One reported message carried six tables, every one of them displayed as `| Metric | Before | After |` followed by `|---|---|---|`, in a proportional font where nothing lines up.

Telegram renders the same content correctly, through either the rich renderer or the HTML fallback. Slack has no table markup at all, so nothing renders it and the raw source is what the user sees.

## Cause

`src/utils/slack_fmt.rs`, `markdown_to_mrkdwn` is a character-level converter: it tracks code fences and inline code and rewrites emphasis markers. It has no concept of a block, so a pipe table is ordinary text and passes through unchanged.

`src/channels/slack/blocks.rs`, `blocks_from_mrkdwn` then splits on horizontal rules and emits `section` blocks plus a `context` footer. Sections take mrkdwn, which has no table support either, so there is no later stage where a table could be rendered.

## What Slack actually supports

Block Kit has no table primitive. The established representation is a preformatted block: monospace, columns padded to a common width, so the alignment survives. That is what other Slack apps do for tabular data, and it is legible where a proportional font is not.

Beyond tables, the block surface is largely unused. Only `section`, `divider` and `context` are emitted today. Available and unused:

- `header` blocks for `#` headings, which currently become bold text inside a section
- `rich_text` with real list elements, instead of literal `-` and `1.` characters
- `fields` for two-column key/value runs, which is what several of the reported tables actually are

## Scope

- Detect GitHub-flavored pipe tables before mrkdwn conversion and render them as aligned monospace inside a fenced block
- Column width from the widest cell per column, so alignment holds
- Respect the separator row's alignment markers where present
- Leave tables inside existing code fences alone
- Do not regress the 3000-char section limit already handled by `split_to_limit`: a wide table must degrade rather than truncate

## Out of scope for the first pass, worth filing separately

Header blocks, rich_text lists, and fields. Those change the shape of every message, so they want their own change and their own before/after review.

## Acceptance

- [ ] A markdown table in a Slack completion renders as aligned monospace, not raw pipes
- [ ] Columns align on the widest cell
- [ ] A table already inside a code fence is untouched
- [ ] A table wider than the section limit degrades without truncating content
- [ ] Non-table content is byte-identical to today's output

adolfousier:
[...]

Issue #876, opened 2026-07-29
fix(telegram): message chunking is markup-unaware, truncating replies at inline code and fence boundaries


adolfousier:
A long reply was delivered cut off mid-sentence at an unclosed backtick: `Why 5760 tests stayed green: there's a test for ` — nothing after it. Reported as recurring ("once again truncated on code blocks").

## Mechanism

`split_message` (`src/channels/discord/handler.rs:24`, reused by Telegram at `delivery.rs:544` and `handler.rs:2511/2541`) splits purely on length:

```rust
let mut end = (start + max_len).min(text.len());
while end < text.len() && !text.is_char_boundary(end) { end -= 1; }
let break_at = text[start..end]
    .rfind(char::from(10))
    .filter(|&pos| pos > end - start - 200)
    .map(|pos| start + pos + 1)
    .unwrap_or(end);
```

It respects UTF-8 char boundaries and prefers a newline within the last 200 chars. It has NO knowledge of markup. So a boundary can land:

- inside `<code>...</code>` or `<b>...</b>`, leaving an unclosed start tag in one chunk and an orphan closing tag in the next
- between a pair of backticks, so the opening backtick renders literally with no partner
- inside a fenced block, splitting it across two messages

Each chunk is sent separately with `parse_mode=HTML`. Telegram parses each independently, so an unclosed tag is a 400 (`Can't parse entities`) and that chunk is lost — which presents to the user as the reply simply stopping.

## Why it is not the truncation-continue path

`looks_truncated_mid_sentence` (`phantom.rs:847`) fires on an alphanumeric final character. A trailing backtick is neither alphanumeric nor in its punctuation list, so that path does not engage. This is delivery-side truncation, distinct from #859.

## Fix

Make chunking markup-aware:

1. Prefer a boundary that is outside any open tag or fence. Track open `<b>`/`<i>`/`<code>`/`<pre>` and backtick/fence state while scanning, and only break where the depth is zero.
2. When no such boundary exists within the window, close the open tags at the end of the chunk and reopen them at the start of the next, so each chunk is independently valid HTML.
3. Never split inside a fenced block if the block fits a single message; prefer breaking before it.

## Test worth pinning

A message whose 4096-boundary falls inside inline code and inside a fenced block must produce chunks that are each independently parseable, and the concatenation must lose no characters.

## Impact

- Long replies silently lose everything after the first bad boundary
- The failure looks like the model stopped, not like a delivery fault
[...]

PR #1608, opened 2026-09-16
feat(discord): live trace, !threads, auto-thread, native tables, utf8 + fence chunk fixes

carvalab:
Implements #1607.

## What

- Tables render natively: narrow ones as an aligned monospace grid in a code fence, wide ones as a Slack-style label layout (#1016 solved the same problem for Slack). Emphasis markers are stripped from cells because nothing renders inside a fence.
- Narration posts into the turn's tool-group bubble as dim subtext, edited in place, instead of one message per paragraph. One work log per turn.
- A message starting with "!" anchors a thread to that message and routes the whole turn into it (tool bubble, trace, answer). DMs fall through, since they have no threads.
- Answers past auto_thread_min_chars post a teaser in the channel and the full body in a thread.
- Thread titles come from a helper: leading bang stripped, whitespace collapsed, cut on a word boundary, marker prefix, so threads are recognizable in the sidebar.
- Fixed mojibake from chunk-split UTF-8: the SSE loop now carries incomplete trailing bytes to the next chunk instead of decoding each chunk in isolation.
- Long messages no longer cut inside open code fences. The chunker widens its search for a clean boundary first so a fence travels whole to the next chunk; when a fence is bigger than one message it closes at the end of one chunk and reopens at the start of the next (#876 tracked the general markup case).

## Config

All opt-in under channels.discord, off by default: trace_narration, bang_new_thread, auto_thread_min_chars. Nothing changes for existing setups.

## Testing

- Unit tests for the table renderer, title helper, UTF-8 carry buffer, and fence-safe chunking, including the upstream markup fixtures
- Full cargo test suite passes
- Verified live against a Discord test bot: thread routing and grid layout. Auto-thread, the UTF-8 carry path, and the fence chunking edge cases are covered by unit tests

Note: this branch's history includes the duplicate-response fix from #1603, since the trace work builds on it.


adolfousier:
Thanks for the work here! One convention note before this can move forward: all tests live under `src/tests/` as dedicated `*_test.rs` files registered in `src/tests/mod.rs` — inline `#[cfg(test)] mod tests { ... }` blocks at the bottom of source files are forbidden by project policy (see **"Where Tests Live"** in `CONTRIBUTING.md`).

You're already doing it right twice in this very PR (`discord_tool_group_test.rs`, `message_split_markup_test.rs`) — the inline blocks just need the same treatment:

- `custom_openai_compatible.rs`
- `discord/handler.rs`
- `discord/table_convert.rs`

Recipe from the docs: create `src/tests/<name>_test.rs`, move the tests over, register the module in `src/tests/mod.rs`, bump any private helpers the tests use to `pub(crate)`, then verify with `cargo test --all-features <name>_test`. Thanks again!

adolfousier:
Merged to main, with thanks @carvalab — two commits:

- `ad84601f` — the six features (live reply tracing, auto-threading, native tables, split fixes + UTF-8 carry). Conflict resolution kept the already-landed #1603 dedup guards; the fence-balance check was adapted to `is_multiple_of` for clippy.
- `ab1c8451` — test extraction per the recipe above: all four inline blocks moved to `src/tests/` (`discord_thread_title_test.rs`, `discord_split_message_test.rs`, `discord_table_convert_test.rs`, `utf8_carry_test.rs`), 17 tests registered, private helpers widened to `pub(crate)`. The four `norm_key` tests were byte-identical to the ones already in `discord_norm_key_test.rs` from #1603, so they were dropped rather than duplicated.

Full gates green (fmt, clippy `-D warnings`, 8796 tests / 0 failed). Closing as completed.

PR #1611, opened 2026-09-16
feat(suggest_options): refuse same-first-word option sets with a rephrase nudge

leshchenko1979:
## Description

Folded long-option buttons render `N. <FirstWord>`; two options sharing a first word would collide (e.g. `1. Run` vs `2. Run`).

This PR adds a tool-layer validation guard in `sanitize_options` (`src/brain/tools/suggest_options.rs`):
- Refuses `n >= 2` option sets where two options share the same case-insensitive first fold word (punctuation stripped).
- Returns a structured `ToolResult::error` nudge naming the colliding word and prompting the model to rephrase so each option starts with a distinctive action verb or keyword.
- Single-option sets are exempt (rendered on the `Go!` tier with no folding ambiguity).
- Updates tool description so models are aware of the uniqueness constraint.
- Includes unit tests covering collision refusal, case-insensitivity, single-option exemption, and distinct option acceptance.

## Verification & CI Gate
- Pre-flight CI Gate: [PR-lane gates run 35088715269](https://github.com/leshchenko1979/opencrabs/actions/runs/35088715269) (formatting, Clippy, and all tests GREEN).
- Live 4-leg smoke verified on running host.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/235


adolfousier:
Merged to main as `5d326aa1` — clean squash, no improvements needed on top. The guard landed exactly as authored:

- `first_fold_word` normalization (lowercase, punctuation-stripped) so `Run` vs `run,` still collide
- HashMap-based duplicate detection over n>=2 sets, with the rephrase nudge in the error
- Single-option exemption (solo renders one button, no fold)
- Tool description updated to teach the model the distinctive-first-word contract
- Existing fixtures adjusted so the cap test trips the cap, not the new guard

5 new tests + updated fixtures, all in `src/tests/suggest_options_test.rs` per project convention — thanks for already doing it right. Gates green (fmt, clippy `-D warnings`, targeted tests 23 passed / 0 failed). Closing as completed.

Issue #165, opened 2026-06-04
In split TUI, strange new session behaviour


leshchenko1979:
I press Ctrl+N, enter a prompt, Tab away to the other pane then Tab back - and I'm back to the previous session that I just ended. 

Have to /session my way back to the new session I just started.

adolfousier:
Seems like the regular behavior. If you create new session and tab back it gets back to the other session that was enabled.

I am not sure I am following. What did you want to do?

leshchenko1979:
Pane A has session X, pane B has session Y.

I am in Pane A with session X.

I press Ctrl+N, start session Z in Pane A.

I enter a prompt, session Z starts rolling. I tab away to Pane B - Pane A freezes. I press Tab again - I switch to Pane A back, but now it's session X in Pane A again. 

Have to /session my way back to session Z in Pane A.

adolfousier:
Confirmed repro. Filed for tracking separately so it does not grow the #163 scope.

Best guess at the path from a quick read: Ctrl+N creates a new session and points the focused pane at it, but on Tab-away the pane retains the old session id from before the new-session creation, so Tab-back restores the previous binding. Will trace the Ctrl+N handler and the pane-session binding on focus return next.

adolfousier:
Fixed on main in 94e592fc.

Two coupled regressions came from the same root cause. The Ctrl+N path updated the in-memory current_session but never touched the focused pane's session_id field or the footer's view of the title.

Pane binding fix. create_new_session in tui/app/messaging.rs now sets the focused pane's session_id to the new session id right after creating it. Mirrors the same sync that load_session already did further down the file. Without this, Tab away and Tab back routed through the is_focus_next_pane handler which loads pane.session_id (the pre-Ctrl+N session), so the new session looked like it vanished and you had to /sessions back to recover it.

Footer title fix. The auto-title spawn in tool_loop.rs writes a generated title to DB via update_session_title but used to never tell the TUI. current_session.title stayed as "New Chat" until the next time load_session re-read the row. Added a lightweight ChannelSessionEvent::TitleUpdated event that fans out through the existing session_updated channel. The TUI handles it with a new TuiEvent::SessionTitleUpdated arm that updates current_session.title and the cached sessions list in-memory. No DB reload, no message reshuffle, no scroll disturbance.

[...]

PR #1615, opened 2026-09-17
fix(brain/tools): advertise the work_status path in tasks_list

leshchenko1979:
## Summary
Advertises the `work_status` file path in `tasks_list` tool output so callers can locate and inspect mid-run subagent execution state without heuristic path guesses.

- Adds `status_file` path reporting to tasks list output formatting
- Updates corresponding unit tests to verify work status path propagation

## Verification
- Pre-flight harvest sweep clean against `adolfousier/main`
- CI Gate Run ID: [35199914165](https://github.com/leshchenko1979/opencrabs/actions/runs/35199914165) (completed success)

Original issue: https://github.com/leshchenko1979/opencrabs/issues/165


adolfousier (commented):
Thanks Alexey. The extraction itself is a cleanup we want: a named, unit-testable `subagent_status_file()` beats an inline `join(format!(...))` at the call site, and it gives the path contract a single place to live.

One thing you should know before this lands, because it changes what the patch is worth here rather than whether we take it: **the defect the PR describes does not exist on this repo's main.** Current `src/brain/tools/tasks_list.rs:122` already resolves through `work_status`:

```rust
let sf = crate::brain::agent::service::work_status::status_dir()
    .join(format!("{id}.json"));
```

and `work_status::status_path` at `src/brain/agent/service/work_status.rs:78` is:

```rust
pub fn status_path(id: &str) -> PathBuf {
    status_dir().join(format!("{}.json", id))
}
```

Byte-identical output. Main has resolved this way since the `#26` P1 unification, so there is no advertised path pointing at the retired `<home>/tmp/subagents` and no ENOENT for the model to hit. The pre-`#26` behaviour the commit message describes is real on your fork, where the linked issue lives, but upstream it was already fixed. Worth flagging so the two histories do not drift in your head: `#165` here is an unrelated issue number.

That has a knock-on for the test. `advertised_status_file_is_the_path_writers_use` asserts `subagent_status_file(id) == work_status::status_path(id)`, and the helper body is a one-line delegation to exactly that call, so it compares a function to itself and passes on pre-patch main too. The second assertion against `legacy_dir()` passes there as well, since the advertised path was already under `detached`.

We are merging with `--no-ff` so your commits stay intact, then stacking on top:

1. **A test that can actually fail.** Instead of comparing the helper to its own resolver, cross the writer boundary: `WorkStatus::new_agent(...)` into a `TempDir` via `test_override`, then read back the string `subagent_status_file()` advertises and assert the file parses with the right kind and session. That pins the real contract (the advertised string points at a file a writer created) and breaks on any directory drift, legacy included. The file already uses this shape at `src/tests/tasks_list_test.rs:50`, so it stays idiomatic.
2. **A doc correction**, dropping the pre-`#26` ENOENT narrative and the fork issue number from the committed comments, replaced with what is true upstream: extracted so the advertised path is unit-testable and resolves through the same `work_status` entry point `src/brain/tools/subagent/spawn.rs:534` writes through.
3. **A `style:` commit** for rustfmt import order, two hunks in `tasks_list_test.rs`. Our CI's fmt step is soft-failed (`continue-on-error: true` at `.github/workflows/ci.yml:104`), which is why it went green on you. That is our pipeline gap, we absorb it.

Nothing for you to do here. Taking the refactor on those terms because the helper earns its place even without the bug behind it.

PR #1617, opened 2026-09-17
fix: reap stale notify_queue rows for unclaimed sessions

leshchenko1979:
## Description

When sessions are terminated, orphaned, or unseated while items remain queued in `notify_queue`, these stale notifications can remain stuck in database queues indefinitely. This fix introduces automated reaping of stale `notify_queue` rows for unclaimed or dead sessions, ensuring queues stay tidy and preventing delivery attempts to non-existent sessions.

## Evidence & Verification

- CI Gate: [`35218685735`](https://github.com/leshchenko1979/opencrabs/actions/runs/35218685735) (`success` in 22m5s)
- Original fork issue: https://github.com/leshchenko1979/opencrabs/issues/182
- Smoke test: Verified in production fleet (`c914464a` / `73bf3f22`).


adolfousier (commented):
Thanks for this one, Alexey. The diagnosis is right and the test suite that comes with it is the good kind: 328 lines, proper `src/tests/` placement, real boundary cases including the empty table. `clear_dead_sessions()` in particular is exactly the reaper this table needed. I checked the migration and `notify_queue` carries no foreign key onto `sessions`, so those rows really are unclaimable forever and nothing else would ever collect them.

We are merging this with `--no-ff` so your commits land verbatim, and stacking our changes as separate commits on top. Here is what those commits do, so nothing is a surprise when you read main.

### The one behavioural change we are making

`reap_stale_unclaimed(cutoff)` is named for unclaimed rows, but the SQL has no claim or route predicate:

```sql
DELETE FROM notify_queue WHERE created_at < ?1
```

Age only. And the call sits above the `repo.all()` redelivery pass, so it runs before anything is offered to a live route. The failure case is a daemon that was simply down for a while: laptop asleep over a long weekend, a machine off across a holiday, a box that sat idle four days. Every parked push older than 72h gets deleted at the next boot instead of delivered, for sessions that are perfectly alive with working routes.

That collides with the invariant written at the top of the module the patch edits (`src/brain/agent/service/notify_queue.rs:14`):

> a DUPLICATE after restart, never a lost message.

Our fix on top is ordering rather than logic: redeliver first, then reap only what is still parked after the pass (rows that re-parked because the route genuinely did not take them), with a regression test that a 72h-old row for a live route gets delivered rather than dropped. Same reaper, same constant, just downstream of the offer so the invariant holds.

### Smaller things we are folding in

- `pub const MAX_ROW_AGE_SECS` is narrowed to `pub(crate)`. Nothing outside the module reads it.
- The `STALE_ROW_SECS` doc says rows past 24h "are logged as defect candidates at boot". No such logging exists in the patch or on main, and the constant only feeds `MAX_ROW_AGE_SECS`, so we correct the comment to match what the code does.
- The `filter_map(... .ok()?)` over reaped rows drops a row with an unparseable `id` or `session_id` silently, while the `bg_meta` branch one line above logs. We add the log so a malformed reap still leaves a trace.
- `now_unix()` is called once per row inside the log loop, snapshotted once instead.
- `cargo fmt --all -- --check` fails on two files, pure import-order churn from a different rustfmt style edition (your order is `database::interact_err` then `Pool`, ours is the reverse). CI did not flag it because the fmt step at `.github/workflows/ci.yml:104` carries `continue-on-error: true`, so this is on our pipeline, not on you. We fix it in a `style:` commit.

None of that is a request for you to change anything. The data-loss ordering is worth knowing about because it generalises: any boot-time cleanup in this module wants to sit downstream of the redelivery pass. The rest is local convention we would rather absorb ourselves than bounce back at you.

Genuinely good catch on the orphan rows. Appreciated.

Issue #1612, opened 2026-09-16
Migrate remaining inline #[cfg(test)] mod tests blocks to src/tests/


adolfousier:
## Context

CONTRIBUTING.md ("Where Tests Live") requires every test to live under `src/tests/` as a dedicated `*_test.rs` file registered in `src/tests/mod.rs`; inline `#[cfg(test)] mod tests { ... }` blocks at the bottom of source files are explicitly forbidden. The rule also says: if you touch a file carrying one, move it as part of your change.

New contributors read source files and copy the local idiom, so every remaining inline block keeps teaching the wrong pattern — recent contributor PRs have added fresh inline blocks into files that had been clean for months. Clearing the backlog removes the ambiguity.

## Current offenders (8 files, 1 block each)

- `src/brain/agent/service/context.rs`
- `src/brain/agent/service/notify_policy.rs`
- `src/brain/timezone.rs`
- `src/brain/tools/skill_gate.rs`
- `src/channels/target_resolver.rs`
- `src/channels/telegram/rich/inline.rs`
- `src/channels/telegram/rich/mermaid.rs`
- `src/tui/render/panes.rs`

Note: `#[cfg(test)]` on test-only helpers/imports (e.g. `src/channels/telegram/governor.rs`, `src/brain/tools/browser/mod.rs`) is the sanctioned helper pattern and is **not** in scope — only the inline `mod tests` blocks listed above.

## Migration recipe (per file, one commit each)

1. Create `src/tests/<module>_test.rs` with the moved tests
2. Register it in `src/tests/mod.rs`
3. Bump private helpers the tests reach to `pub(crate)`
4. Delete the inline block
5. `cargo test --all-features <module>_test` — test count must be identical before and after

Precedent: the May cleanup sweep (2f238227) followed exactly this flow.

## Acceptance criteria

- `grep -rn 'mod .*tests* {' src | grep -v '^src/tests/'` returns nothing (excluding `src/lib.rs` module declaration)
- Full suite green, total test count unchanged

adolfousier:
Done — all 8 legacy inline `#[cfg(test)]` blocks extracted to `src/tests/` per CONTRIBUTING.md.

| Source file | Test file | Tests moved |
|---|---|---|
| `brain/agent/service/notify_policy.rs` | `notify_policy_test.rs` | 1 |
| `brain/agent/service/context.rs` | `agent_service_augment_user_message_test.rs` | 2 |
[...]

Issue #546, opened 2026-07-13
bug(telegram): a no-tool react-only turn silently drops an actionable request (only #439's tools-ran case is guarded)


adolfousier:
## Problem

A turn whose entire output is a `<<react:EMOJI>>` marker with no text AND no tool work is unconditionally accepted as a deliberate react-only acknowledgment, and the bot delivers nothing but the reaction. When the model does this to an ACTIONABLE request (rather than to praise), the request is silently dropped — the bot reacts and stops, the task never runs / never gets an answer. Reported as "if executing a task it reacts and that is it, just stops."

This is distinct from #544 (which only removed the leftover processing-log block on a react-only turn). #544 is a valid UX cleanup but does NOT stop the request from being dropped.

## Root cause

`src/channels/telegram/delivery.rs`:
- `turn_ran_tools = !s.tool_msgs.is_empty()` (~line 142).
- If the final text is empty (react-only) AND `turn_ran_tools` is true → the #439 fallback delivers a "Done — N/M tool calls" summary, so a work turn is NOT dropped.
- If the final text is empty AND `turn_ran_tools` is FALSE → the pure react-only path (the `reaction-only response … skipping text delivery` branch) fires the reaction and returns, delivering nothing.

The false-`turn_ran_tools` branch has no way to tell a genuine praise/ack ("you are one clever mf" → react-only is correct) from a task the model lazily answered with a bare reaction after only reasoning (no tool calls). Both look identical, so the task case is dropped. The tool_loop's "you didn't do the work" guards do not catch it either, because a bare reaction marker does not read as phantom tool-intent — it passes through as a completion.

The 5-minute variant: if the model spends the turn REASONING (streaming, no tool calls) and then emits only a reaction, `turn_ran_tools` is false, so #439 does not apply and the request is dropped.

## Evidence

- On 2026-07-13 the two react-only deliveries (14:33, 14:37) were both 1-iteration, 0-tool turns (`Agent completed after 1 tool iterations` → `delivery.rs:207 reaction-only response`), on the modelscope/Qwen API provider. Those two happened to be praise, but the same branch would drop an actionable request identically.
- The #439 fallback ("executed tools but produced no completion text") fired 0 times on 07-12 and 07-13, i.e. only the tools-ran subcase is guarded and it did not trigger.

## Direction (to design, not prescribed)

[...]

PR #1623, opened 2026-09-17
fix(telegram): prevent false incomplete turn warnings on suppressed deduped text

leshchenko1979:
### Summary

Fixes an issue where Telegram false alarms `⚠️ I ended that turn with only a reaction and did not complete the request...` were emitted on long turns (>= 60s) where the full text had already been successfully delivered via intermediate message bubbles.

### Details
- When response text is streamed or delivered early via intermediate bubbles, `delivery.rs` correctly sets `suppressed_final = true` and clears `text_only = String::new()` to prevent duplicate text from being sent at turn end.
- However, the reaction handler at line 374 evaluated `is_react_only(&text_only)`, which returned `true` on the empty `text_only` string, mistaking the suppressed delivery for an incomplete react-only turn.
- Added a `!suppressed_final` check to ensure suppressed text is not treated as react-only.
- Added regression unit tests covering `is_react_only` and `suppressed_final` dedup behavior.

### Evidence
- Unit tests: `cargo test --lib channels::telegram::delivery` passes.
- CI gate: https://github.com/leshchenko1979/opencrabs/actions/runs/35265941855 (GREEN)
- Live verification on running unit: verified in production on ops profile.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/293


adolfousier (commented):
Thanks for this one. The diagnosis is exactly right and the fix is the minimal form of it.

`delivery.rs:319` already carries `!suppressed_final` on the #546 incomplete-turn notice, and the react-only early return a few lines down was the one place that still read `text_only` without asking why it was empty. Guarding both with the same predicate is the correct symmetry, and the comment you added explains the distinction better than the code above it did.

Two things we are folding in on top rather than asking you to respin.

**1. The inline test module.** `delivery.rs:1239` adds a `#[cfg(test)] mod tests` block. This repo moved every inline block out to `src/tests/` in #1612 and `CONTRIBUTING.md` under "Where Tests Live" now forbids new ones. Nothing you could have known from the diff you branched from, and no reflection on the tests themselves. We are lifting the block to `src/tests/telegram_react_only_dedup_test.rs`.

**2. `test_react_dedup_suppressed_not_react_only` cannot fail.** The body recomputes the predicate from literals:

```rust
let text_only = "";
let suppressed_final = true;
let is_considered_react_only = !suppressed_final && is_react_only(text_only);
assert!(!is_considered_react_only);
```

That is the fix restated in the test, not the fix exercised. It is green with your `delivery.rs` change reverted, so it pins nothing. Your other two tests (`test_is_react_only`, `test_strip_echoed_plan_title`) call real functions and are genuine, so this is one assertion, not a pattern.

The fix on top makes it real by giving the predicate a name the production path also uses:

```rust
pub(crate) fn is_react_only_turn(suppressed_final: bool, text_after_directive: &str) -> bool {
    !suppressed_final && is_react_only(text_after_directive)
}
```

The delivery site calls it, the extracted test calls it, and reverting the guard turns the test red. Same behaviour you shipped, just anchored so a future refactor cannot quietly drop it.

Nothing needed from you here. Merging as is with those two commits stacked on top.

PR #1624, opened 2026-09-17
docs(contributing): explain what clippy catches that cargo check misses

adolfousier:
## Why

CONTRIBUTING.md says "USE clippy, never cargo check" but only states *what*, not *why*. Contributors coming from other ecosystems never learn that clippy's lints are about wrongness, not style, so they treat it as optional polish. This adds the argument, backed by upstream docs.

## What

A "Why clippy, not cargo check" subsection after the lint-pass paragraph:

- clippy reports everything `cargo check` does plus ~700 lints; `clippy::correctness` is the only deny-by-default group and aborts compilation, quoting the official docs ("outright wrong or useless", "carefully picked and should be free of false positives")
- Concrete clean-check-dirties examples, each verified against clippy sources/docs:
  - `clippy::let_underscore_lock` (correctness, `clippy_lints/src/let_underscore.rs`): `let _ = parking_lot lock` drops the guard immediately; the rustc lint only covers std locks
  - `clippy::derive_ord_xor_partial_ord` (deny-by-default): Ord/PartialOrd disagreement breaks BTreeMap/sort invariants silently
  - `clippy::eq_op`: `assert_eq!(a, a)`, the always-passing test
  - `clippy::perf` group: "code that can be written to run faster"
- Even rustc concedes the gap: cargo book states some diagnostics are only emitted during codegen, which check skips
- Local receipt: this repo's own tests carry deliberate `#[allow(clippy::field_reassign_with_default)]` suppressions (`src/tests/onboarding_*_test.rs`), so the gate is real, not decorative

## Proof read

Docs-only change (one paragraph + bullets), no code touched, cargo gates unaffected.

## Related

Follow-up to today's clippy-gate discussion; not closing any issue.


adolfousier:
Merged into `main` as `9b18de72b` (`Merge PR #1624: explain what clippy catches that cargo check misses`), both commits preserved individually rather than squashed, since they are two separate logical changes:

| SHA | Commit | Delta |
|---|---|---|
| [`836d0740`](https://github.com/adolfousier/opencrabs/commit/836d0740b) | docs(contributing): explain what clippy catches that cargo check misses | CONTRIBUTING.md +11 |
| [`5197ec86`](https://github.com/adolfousier/opencrabs/commit/5197ec86f) | docs(contributing): add first-run advisory with clippy --fix path | CONTRIBUTING.md +2 |

Total +13/-0, one file. The new `#### Why clippy, not \`cargo check\`` subsection sits directly under the existing lint-pass paragraph in `CONTRIBUTING.md`, before `### Running the App While You Iterate`, so the section reads as claim then justification then dev loop.

**Receipts checked before merge**, since the text makes factual claims about this repo rather than only about upstream clippy:

- The CI gate quoted is real: `.github/workflows/ci.yml:108` runs `cargo clippy --locked --lib --bins --tests --examples --all-features -- -D warnings`.
- The `#[allow(clippy::field_reassign_with_default)]` suppressions cited exist: `src/tests/onboarding_no_silent_commit_test.rs` (2), `src/tests/onboarding_step_save_test.rs` (2), plus `src/tests/xiaomi_onboarding_test.rs` (2) which the `onboarding_*_test.rs` glob in the text does not cover. The claim stands either way; the lint genuinely fires in this tree and each allow is deliberate.
- Upstream links resolve to the live clippy lint index and the cargo book `cargo-check` page, so the deny-by-default `correctness` framing and the codegen-only-diagnostics gap are both attributable.

CI was green on the head commit before the merge: Build (aarch64-apple-darwin), Test, Lint, Cargo Audit, CodeQL, Analyze (actions/python/rust) and CI Required Gate all passing, Code Coverage skipped. No local sanity run was performed and none was required: the change is documentation only, zero Rust touched, so the clippy and test gates are unaffected by definition.

No issue closed by this PR, it was a standalone documentation improvement.

PR #1628, opened 2026-09-18
fix(cron): mark stuck cron runs as interrupted instead of error

leshchenko1979:
## Problem

`opencrabs doctor --fix` sweeps `cron_job_runs` rows that are still `status='running'` past a max age and stamps them **`error`**:

```sql
UPDATE cron_job_runs SET status='error',
  error='stuck: cleared by doctor --fix (no completion within max age)'
WHERE status='running' AND started_at < ?1
```

The sweep knows only that it saw no completion — not that anything went wrong. For a session-trigger job whose cycle actually succeeded but whose run row was never closed (daemon restart, interrupted turn), the run history therefore shows a **failure**. A healthy cycle becomes indistinguishable from a genuinely failed one, and `doctor --fix` reports `N row(s) marked error` for rows that never failed.

## Change

Introduce `interrupted` as a terminal cron-run status meaning *"no completion observed within max age"*, and have the watchdog/doctor sweep write that instead of `error`.

| File | Change |
|---|---|
| `src/cli/doctor_fix.rs` | Sweep writes `status='interrupted'` and `error='interrupted: cleared by doctor --fix (no completion within max age)'`; `FixReport` detail now reads `N row(s) marked interrupted`. |
| `src/db/models.rs` | `CronJobRun::status` doc comment lists `interrupted` alongside the existing statuses. |
| `src/db/repository/cron_job_run.rs` | Documents the terminal status set on `complete_error`. |
| `src/tests/doctor_fix_test.rs` | Test renamed to `test_clear_stuck_cron_runs_marks_interrupted`; asserts the stale row lands on `interrupted` carrying the new message, and that a fresh `running` row is left untouched. |

`success`, `delivery_failed` and `error` are unchanged: a run that genuinely fails still records `error`. Only the sweep's "we do not know what happened" verdict changes, so any consumer matching `error` continues to see real failures only. The sweep predicate (`WHERE status='running' AND started_at < ?1`) is unchanged, so which rows are swept is unaffected — only the status they land on.

## Verification

Behavioral probe on a build of this branch with an isolated `HOME` and an on-disk SQLite database seeded with two rows:

```
aged row  (started 2h ago)  running -> interrupted
  error = interrupted: cleared by doctor --fix (no completion within max age)
  completed_at stamped
fresh row (started now)     running   (untouched)
report: [stuck-cron-rows-cleared] 1 row(s) marked interrupted
```

CI gate — run https://github.com/leshchenko1979/opencrabs/actions/runs/35293596307 : `Check formatting` ✅ · `Clippy` ✅ · `Run tests` ✅ (GREEN, head `f511c8ed1823f33b8c1919280f5da2d3022da8d8`).

Original issue: https://github.com/leshchenko1979/opencrabs/issues/264


adolfousier (commented):
Good call, and verified safe before saying so.

The concern with a status rename is always the consumers. Checked all of them:

- `cron_job_runs` has no CHECK constraint on `status`, so the new literal is accepted by the schema as written.
- No production code switches on the `error` literal for cron runs. The only writer is `complete_error` at `cron_job_run.rs:89`, and the only reader that renders it (`morning-recap.py:203`) prints the string through, so `interrupted` displays correctly without a change.
- The test update is real: it asserts both the status and the error text, and keeps the live-row-untouched assertion.

Separating a watchdog sweep from a genuine job failure is the right distinction. A stuck row cleared by `doctor --fix` never failed, and calling it `error` made every aged row look like a defect.

One thing we are correcting on top.

The doc block you added to `complete_error` lists `delivery_failed` as a terminal status. Nothing in this repo ever writes it. The only mention anywhere is a test doc comment at `oc_send_scope_test.rs:9` claiming `deliver_result` records it, and the two delivery call sites at `scheduler.rs:838` and `scheduler.rs:866` are both `let _ = deliver_result(...)`, so the outcome is discarded and no status is written at all. The claim predates your PR, but your PR is the one that formalises the status list, so it is the right place to make the list match the code.

We are trimming the documented set to the three statuses that are actually written (`running`, `success`, `error`) plus your new `interrupted`, and filing the dropped delivery outcome separately, since making `delivery_failed` real means handling those two discarded results and that is its own change.

Nothing needed from you. Merging with that doc correction stacked on top.

Issue #265, opened 2026-07-02
Channel commands: fallback @handle strip discards command arguments


adolfousier:
## Finding

\`handle_command\` in \`src/channels/commands.rs\` has a defense-in-depth strip for the \`@botname\` suffix Telegram appends to commands in groups. It cuts the input at the FIRST \`@\` in the whole line and drops everything after it:

- \`/respond_to@yourbot mention\` becomes \`/respond_to\` (the mode argument is lost, so the command shows the current mode instead of switching it)
- any command whose argument legitimately contains \`@\` gets truncated at that point

The primary strip in the Telegram handler (which removes only the bot's own \`@handle\` and keeps arguments) usually runs first, so the broken fallback only triggers when \`bot_username()\` is unknown, but when it does trigger it silently changes the command's meaning.

## Fix

Strip the \`@handle\` from the command token only (the first whitespace-separated word), preserving all arguments. \`/cd@yourbot /some/path\` must become \`/cd /some/path\`.

## Acceptance

- \`/respond_to@yourbot mention\` switches the mode (with argument intact)
- arguments containing \`@\` are untouched
- bare commands and DM commands behave as before

Related: #264

adolfousier:
Fixed in f60f26e7.

The old code cut the command at the first @ sign and threw away the rest of the line. So a command like /respond_to@yourbot mention lost the word "mention" and only showed the current mode instead of switching it.

Now there is a small helper called strip_command_handle. It only removes the @botname part from the command word itself and keeps all the arguments. So:

- /respond_to@yourbot mention becomes /respond_to mention
- /cd@yourbot /tmp becomes /cd /tmp
- arguments with @ in them (like emails) are left alone
- normal text is not touched

Covered by 10 new tests in src/tests/command_handle_strip_test.rs. Full test suite passes.

Issue #247, opened 2026-06-28
follow_up_question buttons land in wrong forum topic


leshchenko1979:
## Summary

`follow_up_question` sends its question with buttons to the wrong forum topic (usually General) when triggered from a Telegram forum topic.

## Root Cause

Two bugs, same symptom:

**Bug 1 — Messages never stored in `channel_messages` table**

`handle_message()` in `handler.rs` receives `channel_msg_repo: &ChannelMessageRepository` but never calls `insert()`. The repo is only used for `update_content()` (agent.rs:210) on edited messages — which updates nothing because nothing was ever inserted.

`ChannelMessage::new_empty()` (db/models.rs:501) has a `with_thread()` method specifically for this, but nothing calls it from the Telegram handler.

**Bug 2 — follow_up_question can not resolve the thread_id**

`follow_up_question.rs:99` → `latest_thread_id_for_chat(chat_id)` → queries `channel_messages` for the most recent message → table is always empty → returns `None` → buttons sent without `message_thread_id` → land in General/current topic.

Even with Bug 1 fixed, there is a **first-message-in-topic race**: the first message in a new topic will not have a DB record yet when `follow_up_question` fires synchronously during the same agent turn.

**Bonus:** `make_approval_callback` (handler.rs:5110) has the same problem — calls `bot.send_message(ChatId(chat_id), ...)` directly with no thread_id.

## Fix required

1. **handler.rs `handle_message()`** — insert the incoming message into `channel_messages` capturing `msg.thread_id` via `ChannelMessage::new_empty().with_thread()`
2. **follow_up_question.rs** — resolve thread_id from the session ID instead of the DB lookup. `TelegramState` already has a `chat_sessions: HashMap<(chat_id, topic_id), session_id>` map (mod.rs:203). Add a `session_topic: HashMap<Uuid, Option<i32>>` alongside `session_chats` so follow_up_question can get thread_id directly — no DB, no race.
3. **handler.rs `make_approval_callback`** — same thread_id resolution fix.

## Additional issue

When `follow_up_question` hits a Telegram API rate limit (HTTP 429 with `retry_after`), it logs `Retry after Ns` and fails instead of retrying after the wait. The timeout should be respected and the send retried.

## Reproducer

1. Create a Telegram group with forum topics
2. Send a message in a non-General topic
3. Agent calls `follow_up_question` in response
4. The question with buttons appears in General or a random topic, not the one the user was in

Issue #233, opened 2026-06-26
bug: Channel hint prefix leaks to user-visible chat


leshchenko1979:
## Bug: Channel hint prefix leaks to user-visible chat

The `[Channel: Telegram — your text response is automatically sent to this chat...]` prefix injected into `agent_input` can appear in the user's Telegram chat. This is internal LLM context that should never be user-visible.

### Code path

**`src/tui/app/handler.rs:2341`** — Telegram handler prepends the channel hint to `agent_input`:

```rust
let agent_input = format!(
    "[Channel: Telegram — your text response is automatically sent to this chat. \
     Do NOT call telegram_send to deliver your answer. Only use telegram_send for: \
     sending to a different chat_id, media, polls, buttons, reactions, or moderation.]\n{agent_input}"
);
```

This prefixed string is what the LLM sees as the user's message.

### Why it leaks

1. `display_text` (clean, no prefix) is correctly passed separately for DB/TUI storage — `tool_loop.rs:737` uses `display_text_override` when present
2. **BUT:** there is no output sanitization to strip the channel prefix from the LLM's response before sending it to the user
3. The LLM sees the prefix as part of the user's message and may echo or reference it in its response
4. The response text is sent to Telegram as-is

### Scope

Same pattern exists for all channels:

| Channel | Line |
|---------|------|
| Telegram | `handler.rs:2341` |
| Discord | `handler.rs:576` |
| Slack | `handler.rs:1284` |
| WhatsApp | `handler.rs:784` |

### Expected behavior

The channel hint should be invisible to the user. It's internal context for the LLM only.

### Suggested fix

Either:
- Strip the `[Channel: ...]` prefix from LLM output before sending to the channel
- Move the channel hint to a system message rather than prepending it to `user_message`

PR #1629, opened 2026-09-18
feat(cron): trigger-gated cron pipeline with mechanical pre-flight and goal dispatch (#233)

leshchenko1979:
## Problem

A cron job in OpenCrabs is **unconditional**: every tick, the scheduler runs the job's agent turn. There is no way to ask *"should this job run now, given the state of the world?"* — the cron expression is the only lever.

That makes a whole class of jobs awkward. A job that matters only when something is actually wrong — a roster sweep that is interesting only when it finds `ORPHAN`/`DETACHED` rows, a health probe that is interesting only on a non-zero exit — must either run its expensive agent turn on every tick and have the agent conclude there was nothing to do, or accept a much coarser schedule. The first burns tokens to reach a conclusion a shell command already reached; the second misses the event entirely.

Three smaller gaps sit alongside it:

1. A job can inject a prompt into a destination session, but it cannot register that prompt as the session's **active goal**.
2. `set_goal` and `deliver_to` were not validated against each other, so `set_goal` could be set alongside a *passive channel* delivery target — which has no session to set a goal in. The job is accepted, and the goal is silently dropped at fire time.
3. Session delivery targets had two spellings in the wild: `bake_delivery_target` wrote `oc://session/<uuid>`, while the scheduler read `session:<uuid>`. A target baked by the former could be refused by the consumer that reads it.

### 1. Mechanical pre-flight trigger gate (`src/cron/trigger.rs`, new)

A job may declare a **trigger command** — a shell command run *before* the agent turn. Its output and exit status are evaluated against a **trigger condition**; if the condition is not met, the agent turn never runs and the tick is recorded as a zero-token run.

| Condition | Fires when |
|---|---|
| `non_empty` *(default)* | stdout/stderr contains any non-whitespace output |
| `exit_non_zero` | the command exits non-zero |
| `exit_zero` | the command exits zero |
| `regex:<pattern>` (alias `re:`) | combined output matches the pattern |
| `always` | the command finishes, whatever its output/exit status |

`TriggerCondition::parse` accepts case-insensitive aliases (`exit_non_zero` / `exitnonzero` / `non_zero`, `exit_zero` / `exitzero` / `zero` / `exit_0`); anything unrecognised falls back to `non_empty`, so an unknown value degrades to the conservative default rather than firing unconditionally. An invalid regex logs a warning and does **not** fire — a typo'd pattern can never silently turn into "run every tick".

`TriggerRunner` executes the command with a timeout and captures stdout, stderr and exit code into a `TriggerResult`.

### 2. Pipeline execution and goal dispatch (`src/cron/pipeline.rs`, new)

`PipelineExecutor` owns the gated path:

- `evaluate_trigger` — runs the pre-flight and resolves it to a `TriggerOutcome`.
- `record_skipped_run` — on a non-firing trigger, records a completed run carrying `Trigger condition not met (exit_code=…, output_bytes=…). Execution skipped.` with **0 input tokens, 0 output tokens, 0 cost**. The tick is visible in run history as a *skip*, not a failure and not an invisible no-op.
- `maybe_dispatch_goal` — when `set_goal` is set and the delivery target resolves to a session, registers the (template-interpolated) content as that session's active goal.
- `interpolate_template` — expands `{output}`, `{stdout}`, `{stderr}`, `{exit_code}` into the goal template.

### 3. Schema and surfaces

```sql
ALTER TABLE cron_jobs ADD COLUMN trigger_cmd TEXT;
ALTER TABLE cron_jobs ADD COLUMN trigger_on TEXT DEFAULT 'non_empty';
ALTER TABLE cron_jobs ADD COLUMN set_goal INTEGER NOT NULL DEFAULT 0;
ALTER TABLE cron_jobs ADD COLUMN goal_template TEXT;
```

`cron add` / `cron update` (CLI) and `cron_manage` (tool) gain `--trigger-cmd`, `--trigger-on`, `--set-goal`, `--goal-template`, with matching schema descriptions.

### 4. Parameter validation (`#247`)

`set_goal` now **requires** a session delivery target (`oc://session/<uuid>` or legacy `session:<uuid>`). Setting it alongside a passive channel delivery is refused at both surfaces, on create *and* on update (validated against the *effective* post-patch `deliver_to`, not just the patch field):

```
Cannot create job: set_goal requires oc://session/<uuid> delivery target.
Channel deliveries (telegram, discord, etc.) are passive outputs and cannot set session goals.
```

A prompt-or-trigger-command presence check is enforced in the same place: a job must have at least one of `--prompt` or `--trigger-cmd`.

### 5. Delivery-target normalization (`#265`)

`bake_delivery_target` normalizes session targets to the `session:<uuid>` wire format the scheduler actually reads, while still *accepting* both spellings on input. Resolution happens once at create/update time, so the concrete target is baked into the job rather than re-resolved on every tick.

## Files

| File | Change |
|---|---|
| `src/cron/trigger.rs` | **new** — `TriggerCondition`, `TriggerResult`, `TriggerRunner` |
| `src/cron/pipeline.rs` | **new** — `PipelineExecutor`, `interpolate_template`, `TriggerOutcome` |
| `src/cron/mod.rs` | re-exports the new modules |
| `src/cron/scheduler.rs` | runs the gate before the agent turn; records the skip |
| `src/brain/tools/cron_manage.rs` | new fields + `set_goal`/`deliver_to` validation |
| `src/cli/cron.rs` | new flags + matching validation |
| `src/channels/target_resolver.rs` | session-target recognition, `is_session_target` / `extract_session_target` |
| `src/brain/tools/subagent/notify.rs`, `src/a2a/handler/notify.rs` | converge `session_notify` goal dispatch |
| `src/db/models.rs`, `src/db/repository/cron_job.rs`, `src/db/database.rs` | the four new job columns |
| `src/migrations/20260915000001_add_cron_trigger_pipeline.sql` | **new** — schema migration |
| `src/tests/cron_trigger_pipeline_test.rs`, `src/tests/cron_delivery_outcome_test.rs`, `src/tests/cron_session_target_test.rs`, `src/tests/a2a_session_notify_test.rs`, `src/tests/cron_test.rs` | coverage for the gate, goal dispatch, validation and target normalization |

## Verification

**Behavioral probe on the live binary** (deployed build `917276229ffdbdfa805d9f8f9daf65f1601a6b44`, running `opencrabs-ops` PID 649951, exe sha256 `07b561577b421b44d339577b1fb2b34292a7ba87cc6917678d3ab9faf8ad9919`).

The trigger gate is exercised on every scheduler tick by a real deployed job, `oc-roster-detached-sweep`:

```
trigger_cmd = oc-roster --classify 2>/dev/null | grep -E 'ORPHAN|DETACHED|MISMATCH' || true
trigger_on  = non_empty
set_goal    = 0
```

When the sweep finds nothing, the pre-flight produces no output, the condition does not fire, and the agent turn is skipped — logged and recorded as a zero-token run:

```
src/cron/scheduler.rs:434: Cron job 'oc-roster-detached-sweep' trigger condition not met — skipping execution
src/cron/pipeline.rs:84:   Cron job 'oc-roster-detached-sweep' trigger skipped — recorded 0-token run
```

Eight consecutive live ticks (01:46:00Z → 02:20:56Z, every 5 min) produced run rows with `status='success'`, `content='Trigger condition not met (exit_code=0, output_bytes=0). Execution skipped.'`, `input_tokens=0`, `output_tokens=0`, `cost=0.0`:

| run | created (UTC) | status | tokens in/out | cost |
|---|---|---|---|---|
| `87260a86` | 2026-09-18T02:20:56 | success | 0 / 0 | 0.0 |
| `4be9d758` | 2026-09-18T02:15:54 | success | 0 / 0 | 0.0 |
| `f8f77dde` | 2026-09-18T02:10:33 | success | 0 / 0 | 0.0 |
| `c15eb938` | 2026-09-18T02:05:25 | success | 0 / 0 | 0.0 |
| `dc575202` | 2026-09-18T02:00:21 | success | 0 / 0 | 0.0 |
| `19095d5a` | 2026-09-18T01:55:18 | success | 0 / 0 | 0.0 |
| `d537154d` | 2026-09-18T01:50:03 | success | 0 / 0 | 0.0 |
| `b35386ce` | 2026-09-18T01:46:00 | success | 0 / 0 | 0.0 |

The live schema carries the migration's four columns (`trigger_cmd`, `trigger_on` defaulting to `non_empty`, `set_goal`, `goal_template`), and `opencrabs cron add --help` exposes `--trigger-cmd`, `--trigger-on`, `--set-goal` and `--goal-template`.

**CI gate** — run [35296780981](https://github.com/leshchenko1979/opencrabs/actions/runs/35296780981): `PR-lane gates (4ab729e9bc60dc3f481c1125bf30a20439065407)` ✅ GREEN (`Check formatting`, `Clippy`, `Run tests` — full test leg executed, not FAST). Head `4ab729e9bc60dc3f481c1125bf30a20439065407` against base `5fc191e46e839c13166c9661ff9d083c83fe5a63`.

Original issues:
- https://github.com/leshchenko1979/opencrabs/issues/233
- https://github.com/leshchenko1979/opencrabs/issues/237
- https://github.com/leshchenko1979/opencrabs/issues/247
- https://github.com/leshchenko1979/opencrabs/issues/265


adolfousier (commented):
This is the largest thing in the current batch and the design holds up. Gating execution on a cheap shell probe turns a polling job into an event job, and recording the skip as a 0-token run instead of silently doing nothing is the detail that makes it operable: you can still see the job ran and decided not to fire. `TriggerCondition` covering exit status, output presence and a regex is the right axis set, the `regex:` / `re:` prefix keeps the common cases unambiguous, and threading the whole thing through `CronJobPatch` with `push_opt` means partial updates behave.

The migration is correct and appended last per the list invariant in `database.rs`, the INSERT was widened to match, and `from_row` defaults every new column so pre-migration rows read back cleanly.

Five things we are folding in on top.

**1. A timed-out trigger orphans its process.** `TriggerRunner::run` wraps `child.wait_with_output()` in `tokio::time::timeout`. On expiry the future is dropped, and `tokio::process::Child` does not kill on drop by default, so the shell and everything it spawned keeps running unsupervised. A trigger that hangs once per schedule accumulates processes for as long as the daemon lives. Adding `.kill_on_drop(true)` to the builder.

**2. `TriggerCondition::as_str` is dead and lossy.** Nothing calls it, in production or in tests. It also does not round-trip: for `Regex(p)` it returns the bare pattern, which `parse` reads back as `NonEmpty` because the `regex:` prefix is gone. Nothing is broken today because `trigger_on` is persisted as the user's raw string and parsed on read, but a future caller would silently degrade every regex trigger to output-presence. Deleting it rather than fixing it, since this repo removes unused code instead of annotating it.

**3. `record_skipped_run` does not record a skip.** The doc says `status="skipped"` and the body calls `complete_success` (`pipeline.rs:61` and `:82`), so a skipped run is stored as `success`. The content string explains it to a human reader, but nothing can query for it, and #1628 landing alongside this makes cron statuses meaningful precisely so they can be queried. Making the doc match the call for now, and noting the real `skipped` status as follow-up rather than inventing a fourth status in a merge commit.

**4. Inline test modules.** `trigger.rs:140` and `target_resolver.rs:371`. #1612 swept every one of these out of production files and `CONTRIBUTING.md` now forbids new ones. Easy here: `src/tests/cron_trigger_pipeline_test.rs` already carries the same assertions, so the inline block in `trigger.rs` is largely a duplicate. Lifting both.

**5. `FORK (#233)` in the migration comment.** `database.rs:105`. In this repo #233 is "bug: Channel hint prefix leaks to user-visible chat", closed and unrelated, so the comment points a future reader at the wrong issue. Rewriting it to describe what the migration does without a cross-repo issue number.

Minor, not blocking and not being changed: `should_fire` recompiles the regex on every evaluation, and `parse` silently falls back to `NonEmpty` for an unrecognised string, so a typo in `trigger_on` quietly changes the condition rather than surfacing.

Nothing needed from you. Merging with those commits stacked on top.

PR #1630, opened 2026-09-18
feat(telegram): process-wide proactive rate limiter and global 429 cooldown lock

leshchenko1979:
### Summary

Adds a **process-wide** rate governor for Telegram sends, closing the gap where per-chat pacing left the bot exposed to Telegram's *global* ~30 req/s limit and to bot-wide 429 flood penalties.

- **Global token-bucket pacer** (`GLOBAL_PACER`, capacity 25, refill 25 req/s, max hold 5s) that every send path acquires before hitting the API, so the aggregate request rate across *all* chats stays comfortably below Telegram's global ceiling.
- **Global 429 cooldown lock** (`GLOBAL_COOLDOWN`): when *any* endpoint returns a 429, every send path observes the lock and pauses for the server-declared `retry_after` window (plus margin), instead of each chat rediscovering the flood independently.
- **Fast-path drop for drop-eligible operations**: typing indicators (`admit_chat_action`) and cosmetic intermediate edits (`edit_admission_media_kb`) drop immediately while a global cooldown is active, rather than spending retry attempts inside an active flood window.
- **Case-insensitive `parse_retry_after`** so servers emitting `Retry-After` with differing case are honoured.

### Why

Per-chat governors cannot bound the *process-wide* request rate: N chats each individually under their own limit can still collectively exceed Telegram's global budget and trigger a bot-wide 429 penalty. This change makes the aggregate request rate a first-class, enforced property of the process.

### CI & Validation

- **CI gate (full, no `--fast`)**: run [`35302740761`](https://github.com/leshchenko1979/opencrabs/actions/runs/35302740761) — job `PR-lane gates (b42f51aa440bccd030873cd081c596c8ae72ab3f)` => success. The job name pins the exact filed sha.
- **4-leg smoke PASS** on the running production daemon, recorded in the canonical `smoke-verdicts.log` (row `2026-09-18T04:00:16Z`, sha `b42f51aa4`):
  - *Lineage* — the port carries the feature; the shipped tree carries the same markers.
  - *Identity* — `oc-smoke-evidence` VERDICT IDENTITY-MATCH: `exe_sha256 == artifact_sha256`, `source_sha == deployed.sha`.
  - *CI gate* — the run above, on the exact filed sha.
  - *Behavioral* — the global-cooldown seam fired on the live daemon after the 03:52:34Z swap (2 activations from the current binary, 542 across the day).

Original issue: https://github.com/leshchenko1979/opencrabs/issues/262


adolfousier (commented):
This is the right shape for the problem. A per-chat pacer cannot see a process-wide ceiling, and a 429 discovered in one chat is information every other chat needs. The monotonic `max(existing, now + wait + margin)` extension is the correct semantics for concurrent 429s, the fail-open at `GLOBAL_MAX_HOLD` is the right instinct for a pacer that must never hang a turn, and routing the fast-path drop through `is_global_cooldown_active` for typing refreshes and non-final edits keeps cosmetic traffic from queueing behind the lock. Lowercasing in `parse_retry_after` is a good catch on its own.

Three gaps we are closing on top.

**1. The global ceiling is disabled by the per-chat switch.** In both pacers the new call sits after the per-chat enable gate:

```
governor.rs:987   if !lim.enabled { return; }
governor.rs:992   acquire_global_permit().await;
```

and the same order at `:1079` / `:1084`. Turning off the per-chat limiter therefore also removes the process-wide ceiling and the 429 cooldown wait, which is the opposite of what a global limiter is for: the per-chat setting is a per-chat policy, the global one is a safety floor. We are hoisting `acquire_global_permit()` above the `lim.enabled` check in both.

**2. `record_global_429` is armed from one site only.** It is called exclusively from `wait_out` (`rate_limit.rs:167`). Several production paths recognise a 429 and never reach it, most clearly the plan card:

- `plan_card.rs:284` (edit failure) and `plan_card.rs:303` (create failure) both parse a real retry-after, suppress card writes locally, and return without touching the global deadline.

So a 429 discovered on a plan-card write teaches the rest of the process nothing, which is the exact blind spot this PR exists to remove. We are adding `record_global_429(wait)` at both, and auditing the remaining direct `status == 429` sites in `rich/api.rs` and `send.rs` in the same commit.

**3. Worth knowing rather than fixing.** `GLOBAL_PACER` is a `std::sync::Mutex` taken inside an async fn. The guard is dropped before the `.await` in every path here, so it is correct as written, but it is the kind of thing that breaks silently if someone later awaits inside that block. We are leaving your code alone and adding a note at the lock.

The two test files are well built, particularly scripting the virtual clock through `test_support::advance` so the cooldown wait is deterministic rather than a real sleep. Nothing needed from you. Merging with the fixes stacked on top.

Issue #782, opened 2026-07-26
Telegram: /new is not owner-gated, and the command menu is one global list


adolfousier:
Two related gaps, one at execution and one at display.

## 1. `/new` has no owner gate

`commands.rs:449`:

```rust
"/new" => ChannelCommand::NewSession,
```

Every sibling command carries `if !is_owner { UnknownCommand("🔒 Owner-only command.") }`. `/new` does not, so any user in an allowlisted group can start a fresh session and reset the owner's context.

## 2. The menu is a single global list

`register_bot_commands` calls `bot.set_my_commands(commands)` with **no scope**, so every user of the bot sees the same ~22 commands. Since every command except `/new` is already owner-gated at execution, a non-owner sees a full menu where every entry answers "🔒 Owner-only command."

`/start` has the opposite problem: it is shown to everyone including users already on the allowlist, for whom it is meaningless — they are already registered.

## Wanted

Scope the menu to who is looking:

| audience | menu |
|---|---|
| owner | the full list, including `/new` |
| not in the allowlist | `/start` only — the one command they need |
| in the allowlist, not owner | nothing — every command is owner-only anyway |

teloxide 0.13 exposes the scopes needed: `Default`, `Chat { chat_id }` and `ChatMember { chat_id, user_id }`.

## Notes

The registration currently runs once at startup. Per-member scoping means it also has to run when the allowlist changes, or allowed users keep whatever menu they were last given.

Related: #776 (`/start` gate misleads already-allowed users). Same command, different half of the problem — that one is about the reply text, this one is about menu visibility.

adolfousier:
Both done in `34558b5d`.

## `/new` owner gate

```rust
"/new" => {
    if !is_owner {
        ChannelCommand::UnknownCommand("🔒 Owner-only command.".to_string())
    } else {
        ChannelCommand::NewSession
    }
}
```

Two tests, matching the `/cd` gate pair already in `channel_command_owner_gate_test.rs`.

## Menu scoping

| audience | menu |
|---|---|
| default | `/start` alone |
| owner | full list, in DMs and in every configured group |
| allowed non-owner | empty |

`/start` also leaves the owner's own list, since the owner is on the allowlist by definition and has nothing to register.

Uses `BotCommandScope::Default`, `Chat { chat_id }` for the owner's DM, and `ChatMember { chat_id, user_id }` per group — for the owner to grant the full list, for everyone else already allowed to clear it.

## Known limitation, not solved here

[...]

PR #1631, opened 2026-09-18
fix(telegram): publish default group command scopes and wire my_chat_member for group & forum menu visibility

leshchenko1979:
## Problem

When the bot is added to a plain group, a supergroup, or a Telegram forum, members
see **no command menu** — only the `/start` floor, or an empty command picker.

### Root causes

1. **No group-level baseline scopes.** `register_scoped_menus` publishes only
   `BotCommandScope::Default` (a `/start` floor) plus a `BotCommandScope::Chat`
   entry for each group explicitly listed in `config.toml`. Any group the operator
   did not hand-configure inherits the `/start`-only default.

2. **`my_chat_member` was a no-op for registration.** The bot's own
   membership-change handler only cleared the solo-owner evaluation cache, for
   logging and cache hygiene. It never triggered menu registration.
   `member_events.rs` does fire `maybe_auto_register` off `new_chat_members`
   service messages — but **forums routinely suppress or omit join service
   messages**, so in a forum `my_chat_member` is the only join event the bot ever
   sees and registration never ran.

3. **Solo-owner evaluation refused shared groups.** `evaluate_solo_group` returned
   `SoloEval::OtherHumans(..)` as soon as any human besides the owner was present,
   so the owner's own catalog was never published in any group with other members.

## Fix

Three changes, one cohesive unit:

**1. Publish group baseline scopes at startup** (`src/channels/telegram/agent.rs`,
`register_scoped_menus`):

- `BotCommandScope::AllGroupChats` → `/start`, `/help` for every member of every
  group the bot sits in.
- `BotCommandScope::AllChatAdministrators` → `/start`, `/help`, `/doctor`,
  `/usage` for group admins.

Both are best-effort (`tracing::warn!` on API failure) and additive to the
existing default and per-chat scopes.

**2. Wire `my_chat_member` to registration** (`src/channels/telegram/agent.rs`):
on a transition to `Member` / `Administrator` / `Owner` in a non-private chat,
spawn `menu_auto::maybe_auto_register`, so a freshly added bot registers its menu
without waiting for user message traffic. A second, lazy trigger rides the
inbound-message path: a message from a non-private chat that has never been
evaluated kicks off the same background evaluation. That covers chats where the
bot was already present before this build, or where the membership update was
missed entirely.

**3. Owner presence is the trigger, not exclusivity**
(`src/channels/telegram/menu_auto.rs`): `evaluate_solo_group` now returns
`SoloEval::Eligible` whenever the owner appears in the observed member list;
strangers and extra bots no longer block it. The owner's full catalog is published
under `BotCommandScope::ChatMember { chat_id, user_id: owner }`, so it is visible
to the owner alone and does not leak operator commands into a shared group.
`SoloEval::OtherHumans` is removed, and the `OwnerAbsent` path now records the
negative evaluation so the membership probe is not repeated on every message.

Net effect for a group with no `config.toml` entry:

| Viewer | Menu |
|---|---|
| any member | `/start`, `/help` |
| group admin | `/start`, `/help`, `/doctor`, `/usage` |
| the bot owner | the full command catalog |

## Tests

`src/tests/menu_auto_solo_test.rs` pins the new decision rule: the owner's
presence alone yields `SoloEval::Eligible`, including with other humans and bots
in the member list. No inline `#[cfg(test)]` blocks were added — tests live under
`src/tests/` per CONTRIBUTING.

## Verification

- **CI gate** — full triad (fmt + clippy + `cargo test --all-features`), no fast
  path: run [35348209510](https://github.com/leshchenko1979/opencrabs/actions/runs/35348209510)
  on head `925556491` — **GREEN** (Check formatting, Clippy, Run tests all
  `success`).
- **4-leg smoke on the running binary** (lineage, identity exe-sha == artifact-sha,
  CI gate, live behavioral probe on a swapped build): PASS, recorded in the fork's
  `smoke-verdicts.log`.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/248


adolfousier (commented):
Revised after testing this against a live deployment. The premise does not hold here, so this replaces what I first wrote.

**The owner already gets the full catalog in every configured group on main.**

`register_scoped_menus` publishes the full command list under `BotCommandScope::ChatMember { chat_id, user_id: owner }` for every entry in `channels.telegram.groups`, unconditionally:

```
agent.rs:2594   for group_id in tg.groups.keys() {
                    // Owner inside the group.
                    bot.set_my_commands(commands.clone())
                       .scope(BotCommandScope::ChatMember { .. })
```

There is no member check on that path. Other humans, admins, bots, forum or plain group, it publishes regardless.

The solo-owner rule you relaxed never touches those chats at all. `maybe_auto_register` returns before evaluating anything when the chat is configured:

```
menu_auto.rs:86   if tg.groups.contains_key(&chat_id.to_string()) {
                      return;
                  }
```

So root cause 3 as written, that the owner's catalog "was never published in any group with other members", is not the behaviour of this codebase. It is only true for a group with no `config.toml` entry, and for that case the supported remedy is a config entry, which also brings the per-member clearing and the supergroup-migration follow that this path does not have.

Verified by direct observation, not by reading: on a live deployment with 21 configured groups, the owner sees the full slash-command menu in all of them today, on main, in groups with many other members.

**Second problem: the group baselines misrepresent the permission matrix.**

`AllGroupChats` advertising `/help`, and `AllChatAdministrators` advertising `/help`, `/doctor` and `/usage`, promises commands that refuse the people being shown them. All three are owner-gated at execution:

```
commands.rs:425   "/doctor" => if !is_owner  -> "Owner-only command."
commands.rs:446   "/help"   => if !is_owner  -> "Owner-only command."
commands.rs:568   "/usage"  => if !is_owner  -> "Owner-only command."
```

A Telegram group administrator is not the bot owner. This is the exact state #782 removed, and the doc comment above the function you extended still says so.

**Closing this one.** The relaxation fixes a case that the configured-group path already covers, and the baselines re-open a resolved defect, so there is no net behaviour here worth merging.

If the forum gap is real and separable, it is worth a fresh, focused PR: only the `my_chat_member` wiring, with no baseline scopes and no change to `evaluate_solo_group`, plus a log excerpt from a forum where registration genuinely never ran. Evidence of the failure firing is what would make that one easy to take. The `my_chat_member` idea itself is sound and I would like to see it land on its own.

Thanks for the work on this, and for the reasoning in the description. It was the part about configured groups that I should have checked before agreeing with it.

adolfousier (commented):
Superseded: the review above is the current one. GitHub does not allow deleting a submitted review, so this entry stays as a stub.

Issue #627, opened 2026-07-19
CI: drop the Windows build job from the PR gate (chronic 40m timeout, redundant with release)


adolfousier:
The Windows cross-platform build in ci.yml is a required PR gate but chronically times out at the 40m cap: it builds --all-features (the heaviest chain, incl. local-stt/local-tts) with rust-cache deliberately skipped on Windows (always cold), so it routinely blows past the limit and blocks otherwise-green PRs (e.g. PR #626 — every other check passed; only Windows timed out mid-dependency-compile).

It also adds little: it is build-only (no tests), and release.yml's build-release matrix already builds and validates Windows before any tag is published, so a platform regression cannot reach a release unnoticed.

Fix: remove the windows-latest entry from the ci.yml build matrix (and its now-dead Windows-only steps). Keep the macOS aarch64 build for cross-platform sanity. Linux test + lint + audit + macOS build remain the PR gate; Windows coverage stays where it matters, at release time.

adolfousier:
Done in 827f7e3e (pushed to main). Removed the windows-latest entry from the ci.yml build matrix and its now-dead Windows-only steps; kept the macOS aarch64 build. The PR gate is now Linux test + lint + audit + macOS build. Windows stays validated by release.yml's build-release matrix before any tag is published, so platform regressions still cannot reach a release. Note: the build job is PR/workflow_dispatch-only, so direct pushes to main were never affected.

Issue #1168, opened 2026-08-23
feat(tools): write_file partial-view overwrite guard


adolfousier:
# feat(tools): write_file partial-view overwrite guard

## Problem

`write_file` happily overwrites an existing file with no knowledge of whether the file was **fully read** in this session. A windowed read (`start_line` / `line_count`) followed by a full overwrite silently destroys every line the model never saw. The read-tracking grep (`fully_read`, `read_state`) returns zero matches in `write.rs` — nothing tracks it.

The awesome-opencrabs brain carries a read-before-write prompt rule patching exactly this. Related: #954 is the **cross-agent** half (stale read overwritten by a concurrent agent). This issue is the **intra-session** half: a single agent windowed-reads, then overwrites from a partial view.

## Evidence

- `src/brain/tools/write.rs` — overwrite path has no read-state input; `create_dirs` handling at `:100` is the only pre-write gate
- `read_file` windowed reads (`start_line`, `line_count`) leave no marker of what was seen
- #954 (open) — concurrent stale-read overwrite, same file, different trigger

## Suggested fix

Track fully-read paths in session state. When `write_file` targets an existing file that exists but was **not fully read this session**, require an explicit opt-in:

- New param `overwrite_read_confirm: bool` (default false)
- Without it: error stating file size, what (if anything) was read, and pointing at the confirm param
- New files, fully-read files, and confirmed overwrites proceed as today

Coordination note: whoever picks this up should check #954 first — if that fix introduces per-path read/version tracking, this guard should ride the same state rather than a second mechanism.

## Tests

- Windowed read + overwrite without confirm → error
- Same with `overwrite_read_confirm: true` → succeeds
- New file and fully-read file overwrite without confirm → succeed

cc @alexeyleshchenko — from the brain-files analysis

adolfousier:
Implemented. Commit `b94c9b36` (7 files, +252/-1).

**Design:** process-wide registry keyed by `(session_id, canonical path)` in new `brain/tools/read_state.rs` — tool instances are constructed independently, so per-instance maps would never meet; test isolation falls out of each context's fresh session UUID.

| Path | Behavior |
|---|---|
| New file | Writes straight through, no state consulted |
| Existing file, session fully read it | Unlocked, proceeds as today |
| Existing file, only windowed reads | **Guarded**: error names the size, states the file "was not fully read in this session — a windowed read doesn't count", and offers both exits: full re-read or `"overwrite_read_confirm": true` |
[...]

Issue #954, opened 2026-08-06
fix(tools): concurrent agents silently lose each other's edits — write_file overwrites from a stale read


adolfousier:
Agents running at the same time in one repo overwrite each other's edits, silently. A file is changed, another agent writes it from a stale copy, and the first change is gone with nothing reported to anyone.

## The tool that clobbers is `write_file`, not `edit_file`

`edit_file` re-reads the file inside the call (`edit.rs:214`) and applies a targeted operation to that fresh content. So it is largely safe by construction:

- another agent changed a **different** part → the re-read picks that up and the edit lands on top
- another agent changed **the same** text → `old_text` no longer matches and the tool errors out

Neither case loses data.

`write_file` never reads (`grep -c read_to_string src/brain/tools/write.rs` → **0**). It writes `input.content` wholesale:

```rust
// write.rs:165
fs::write(&path, &input.content)
```

That content was composed by the agent from a read that happened in an **earlier tool call**. Anything written to the file between that read and this write is destroyed, silently. That is the lost update, and it is the whole of it.

## Why this is a shipped default, not an edge case

Every one of these builds its own `AgentService` pointed at the same working directory:

| Site | Writer |
|---|---|
| `cli/ui.rs:1005` | TUI agent |
| `cli/ui.rs:413` | daemon agent |
| `channels/factory.rs:144` | every channel agent — Telegram, Discord, Slack, WhatsApp each get one |
| `tui/app/state.rs:1590` | provider-swap rebuild |
| `subagent/spawn.rs:257` | every spawned sub-agent, `.with_working_directory(context.working_dir())` |
| `subagent/resume.rs:210` | every resumed sub-agent |
| `subagent/team/create.rs:234` | every team member |

`team_create` exists to *"spawn multiple sub-agents at once… as one atomic team"*, each inheriting the parent's directory, and `agent.max_concurrent` defaults to **4**. So four agents writing one tree is the default configuration before anyone opens a second pane or sends a channel message.

## There is no arbitration anywhere

No lock, no file lock, no compare-and-swap, no mtime check on the project-file write path.

Config writes are guarded — `CONFIG_FILE_LOCK` exists precisely because a read-modify-write race there corrupted keys. Project source files have no equivalent.

## Proposed fix: detect the lost update, do not prevent the concurrency

Refusing the write is enough. The agent is perfectly capable of re-reading and redoing its change; what it cannot do is notice that its write silently destroyed someone else's.

[...]

PR #1632, opened 2026-09-18
Windows reliability: shell spawning, cwd, encoding, search fallback

moneyacademyKE:
Fixes 10 verified Windows failure modes, each reproduced on a bare Windows
Server 2022 box (the same class of host the Windows release targets) and
pinned with a regression test where a unit-level pin is possible. Windows is
currently never compiled in CI (#627) and never tested — this PR also adds
the CI compile gate so this class of bug stops shipping.

## The failure modes

1. **`sh -c` hardcoded in four spawn sites** (`background_tasks.rs`,
   `dynamic/tool.rs`, `plan_tool.rs`, TUI `!` bang). On stock Windows `sh`
   does not exist, so every detached background task, dynamic shell tool and
   plan verification gate failed with "program not found". Linux CI passed
   because `sh` exists there. Fix: `utils::shell::shell_pair()` — one source
   of truth for the platform shell pair (`cmd /C` / `sh -c`).
2. **MSVC argument quoting corrupts quoted commands** — `Command::arg()`
   applies MSVC escaping, `cmd.exe /C` re-parses with its own rules, so any
   command containing quotes is mangled (e.g. `python "C:/x/probe.py"`
   receives `C:\Windows\System32\"C:\Users\…"` as argv; `dir "C:\Program
   Files"` fails with "filename syntax incorrect"). Fix:
   `PushShellCommand::push_shell_command` — `raw_arg` (verbatim) on Windows,
   plain `arg` elsewhere; all spawn sites converted.
3. **Timeout kills only `cmd.exe`, orphans the work process** — tokio's kill
   TerminateProcess()es the direct child; the grandchild (cargo, cmake, …)
   survives holding target-dir and package-cache locks for minutes. Fix:
   `taskkill /PID <pid> /T /F` sweep on the timeout error path
   (`kill_process_tree`).
4. **Service-launch cwd is `C:\Windows\system32`** — a daemon started by a
   service/scheduled task/Run key begins in the system directory, so every
   relative tool path resolves there, `write_file` refuses home targets as
   "outside the working directory", and Runtime Info advertises the system
   dir as the project. Fix: `utils::cwd::launch_cwd()` — fall back to home
   when the process cwd is a Windows system directory; nine startup sites
   converted.
5. **`read_file` hard-fails on non-UTF-8 files** — `stream did not contain
   valid UTF-8` for `.edn` logs, `.marker`, event logs, and any PowerShell
   `>` redirect (UTF-16LE with BOM by default). Fix: BOM sniff (UTF-16LE/BE
   decoded in-memory), UTF-8 passthrough, otherwise lossy decode with a
   visible note — never a refusal to read.
6. **`session_search` with an empty query errors instead of listing** — the
   #1 avoidable failure in the feedback ledger (600+ logged failures, mostly
   cron auto-resume loops that treat an optional field as a mandatory
   placeholder). Fix: empty query now falls back to the session list with a
   teaching note.
7. **`~\backslash\paths` not expanded** — models routinely paste
   `~\.opencrabs\logs` with native separators; the `~` stayed literal and the
   path silently resolved relative to the working directory. Fix:
   `expand_tilde` handles the `~\` form.
8. **`findstr` no-match exit 1 surfaced as a tool failure** — no-match is
   benign (grep semantics), and the failure loop teaches the model to retry
   variants. Fix: no-match maps to empty success like `grep`.
9. **"gh is preinstalled" is false on stock Windows** — the bash tool
   description and the system prompt both assert it, sending the model into
   a failing `gh` loop. Fix: both now say "check with `gh --version`, fall
   back to `http_request` against the REST API".
10. **Windows is never compiled in CI** (#627) — the reason all of the above
    shipped. Fix: `windows-build` job in `ci.yml`, included in the required
    `ci-gate` set.

## Rig notes (debug builds on Windows)

Debug builds of this repo fail to link on Windows out of the box:
`esaxx-rs` 0.1.10 compiles `/MT` (its build.rs calls `.static_crt(true)`
unconditionally) while `llama-cpp-sys-2`'s cmake defaults to `/MD` →
LNK2038. Release builds on CI don't hit it the same way, but local debug
builds do. Workaround (rig-local, not committed — see the job comment in
`ci.yml` for the full recipe): set `CMAKE_C_FLAGS` / `CMAKE_CXX_FLAGS` (+
`_RELEASE` variants for the VS generator) to `-MT` before building;
cmake-rs skips its own `-MD` injection when the build script already
defines them, and llama-cpp-sys-2 forwards all `CMAKE_*` env vars to cmake.

## Testing

Measured on a bare Windows Server 2022 box (MSVC 1.98.1, default
features), full `cargo test --lib`, drift-as-delta against the base
commit (base = 8738 passed / 58 failed; this branch = 8756 passed /
54 failed):

- **Zero regressions** — every test passing on base passes here.
- **4 base failures fixed by these changes** (the `sh -c` hardcode and
  the system-dir cwd were breaking them): `spawn_command_enqueues_on_
  completion`, `test_execute_shell_echo`, `coerce_non_empty_value_
  untouched`, `persist_sticky_pair_does_not_write_session_state`.
- **15 new regression tests, all green** (shell pair + verbatim quoting
  + tree-kill, cwd sanitize, BOM/UTF-16/lossy read, session-search
  fallback, tilde backslash, findstr benign exit).
- The 54 remaining failures are **pre-existing Windows-blind tests**
  (Unix commands baked into scripts — `pwd`/`cat`/`true` — systemd,
  Windows-different file-lock semantics, path-separator assumptions).
  This was the first Windows test run in the repo's history; the
  failure list is the backlog for a follow-up.
- Note: two pre-existing Windows-only warnings (`path_lock.rs` unused
  `Instant`, `tui/app/input.rs` unused `Command`/`Stdio` — both used
  only under non-Windows cfg) demonstrate the same "never compiled on
  Windows" gap the CI job closes.

## Commits

Atomic per fix; `utils::shell` + `utils::cwd` land as their own module
commit, the five test files are registered in a final `test:` commit
(single `src/tests/mod.rs` index), CI last.


adolfousier (commented):
Read the whole thing against current main. The engineering is genuinely good: each fix carries a repro, the module docs explain the *why* rather than the *what*, and the commits are already sliced along the right seams. Fixes 1/2/4 (the `sh -c` hardcode, MSVC arg escaping, the `C:\Windows\system32` cwd) are real bugs that only a Windows rig would ever surface, and they survived months of green CI precisely because the gate never compiled Windows.

We are taking it: merging into main locally and landing the follow-ups on top as separate atomic commits, rather than asking you to re-roll 12 commits. Three things are changing in that follow-up work, and one of them is a correctness bug worth your attention.

## 1. The lossy `read_file` silently unlocks `write_file` on binary files

This is the one I would not ship as-is. Chain, all verified in the merged tree:

1. `decode_file_bytes` now returns lossy text for any non-UTF-8 input instead of erroring (`read.rs`, small-file path). The NUL-byte note is good, but it is a *warning*, not a failure.
2. That same path then calls `read_state::mark_fully_read` and `file_versions::record`, which is exactly what arms a wholesale overwrite (#1168, #954).
3. The staleness guard cannot catch it either: `write.rs` does `fs::read_to_string(&path).await.ok()`, which yields `None` on a binary file, and `is_stale_write` returns `false` when `on_disk` is `None` (its own comment says "creating a new file races with nobody").

Failure scenario: a model reads `~/.opencrabs/opencrabs.db` (no extension match in `media_tool_redirect`, under the large-file threshold), gets U+FFFD soup recorded as a successful whole-file read, later calls `write_file` on that path, and both guards pass. The DB is replaced with mangled text. On main today step 1 hard-fails, so nothing downstream ever arms.

The BOM and UTF-16 handling is correct and worth keeping. The fix is narrow: a lossy decode must not count as a whole-file read. We are landing that on top.

Two smaller ones in the same area: the UTF-16 path uses `as_chunks::<2>().0`, which silently discards a trailing odd byte (a truncated file loses its last half-character with no note), and `kill_process_tree` drops both `Command::output()` results with `let _`, so a best-effort sweep that never runs is indistinguishable from one that worked.

## 2. `windows-build` is coming out of the required gate

Not a judgement on the job, a resource decision with receipts behind it. #627 removed Windows from the PR gate on 2026-07-19 for chronically blowing a 40m cap, with release.yml's build matrix keeping Windows validated before any tag ships. This PR puts it back into `ci-gate.needs`, and the measured numbers on your own run are:

- Windows build: 51m09s (the only successful run on the branch; the other attempt was cancelled at 40m14s)
- Gate wall-clock goes from ~43m to ~65m, so +51% on every PR
- 160 PR-triggered CI runs in the last 14 days, which is roughly 136 hours of Windows runner time per fortnight, against 3 Windows builds (~2.5h) that release.yml actually ran in the same window

Also note every other gate job runs self-hosted; `windows-build` is the only GitHub-hosted one in the set, which is why it is the one that drags. And the new 60m cap leaves 15% headroom over the measured 51m, so one dependency bump that busts the cache reproduces the exact #627 failure: every check green except Windows timing out mid-compile.

The bug you are protecting against is real, so the job stays, just not as a blocker: out of `ci-gate.needs`, and paths-filtered so it fires when something Windows-relevant changes (`src/utils/shell.rs`, `src/brain/tools/bash.rs`, `cfg(windows)` sites, `Cargo.lock`) instead of on every push to every PR. Same coverage for the class of bug this PR fixes, roughly an order of magnitude less runner time.

## 3. The prompt change contradicts an always-loaded brain file

The `gh` routing edit tells the model to fall back to `http_request` against the GitHub REST API. The shipped `AGENTS.md` says the opposite in capitals, because `http_request` gets 403 from api.github.com over missing User-Agent support. Your underlying point is correct (`gh` is genuinely not preinstalled everywhere, so "it is preinstalled" was wrong), but the fallback needs to be something that works. We are reconciling that on our side.

## Notes, no action needed

- The Unix half of the tree-kill has no test (`kill_process_tree_terminates_the_tree` is `#[cfg(windows)]`) and was added in a late commit to make the sweep compile, so every macOS/Linux user gets a new `pkill` on every bash timeout that no rig has exercised. We will cover it.
- `session_search` empty-query returning a list instead of an error is a tool-contract change on all platforms, justified by a Windows ledger stat. Reasonable, keeping it, just flagging that it is not Windows-scoped.
- Your own note says 54 Windows test failures remain. Understood as the follow-up backlog: this makes Windows compile-gated, not green.

Thanks for this. The four spawn sites alone were worth the PR.

PR #1633, opened 2026-09-18
fix(brain): accept valid profiles/ paths in write_opencrabs_file

leshchenko1979:
## Symptom

`write_opencrabs_file` refuses a legitimate path whenever it carries a leading `profiles/` segment, even when that segment is exactly what the path needs.

With the active home at the default (non-profile) root, a sibling profile's brain file is reached through a path that is *correct as written*:

```
write_opencrabs_file(path="profiles/family/USER.md")
→ Path 'profiles/family/USER.md' starts with 'profiles/' which looks like you
  included a directory prefix. This tool expects a relative path from your
  home directory.
```

The path is not malformed — `profiles/<name>/…` is the only relative form that addresses another profile, and it does not duplicate the home. The write is refused anyway.

## Root cause

`validate_opencrabs_path` decided the case on the **raw string prefix**, not on the resolved path and not on the shape of the home it was resolving against:

```rust
if path.starts_with("profiles/") {
    return Err(...);
}
```

That single test is wrong in both directions:

1. **False positive.** When the home is the default root (`<base>`), `profiles/<name>/…` is a real, non-duplicating target — and it was rejected.
2. **Bypass.** The check is component-blind, so `./profiles/ops/TOOLS.md` never matched the literal prefix and slipped through — accepted where the prefix genuinely *does* duplicate a named-profile home.

The two errors are the same defect: a string test standing in for a path test.

## Fix

Gate the refusal on **the home's own shape**, and make the path test **component-based**:

- `home_is_named_profile(home)` — true only when the home itself is `<base>/profiles/<name>`, i.e. the case where a leading `profiles/` segment really does duplicate the home. This is what makes the refusal correct for a named-profile home and absent for the default root.
- `leading_profiles_segment(path)` — walks `Path::components()` and skips `CurDir`, so `./profiles/…` and `profiles/…` get the same verdict. The bypass closes.
- The refusal now fires **only** when both hold, and the error names the resolved home instead of asserting the path is malformed.

`validate_opencrabs_path` takes the home as a parameter so the decision is testable against any home shape, and the call site resolves the home once.

## Tests

`src/tests/brain_tools_write_opencrabs_file_tests.rs` (registered in `src/tests/mod.rs`; no inline test modules in `src/`):

- the two predicates directly: named-profile vs default home, and `profiles/…` / `./profiles/…` / bare `profiles` / `MEMORY.md` / `memory/profiles/note.md` / empty string;
- valid named-profile paths accepted;
- valid default-profile `profiles/<name>/…` paths **accepted** (the regression this fixes);
- duplicate-prefix paths **rejected**, including the `./profiles/…` form that previously bypassed the check;
- absolute, `~`, null-byte and traversal paths still rejected.

## Evidence

**CI — full gate, no `--fast`** (fmt + clippy + tests, 100% upstream parity):
https://github.com/leshchenko1979/opencrabs/actions/runs/35393512152 — conclusion `success`, job `PR-lane gates (7c6747b42ec2962580801e69f9a05a8c15bb2246)`.

**Behavioral probe on the running binary** (three arms, read from the delivered artifact):

- *accept arm*, temp default-profile home, before/after on the same probe: pre-fix the call was REFUSED with the old message and the target file never landed; post-fix the same call returns success and writes the file (16 bytes).
- *refuse arm*, live named-profile home: `profiles/<name>/…` refused with the new message naming the resolved home.
- *bypass arm*, live: `./profiles/<name>/…` — accepted before the fix — is now refused.

Zero stray files were created in the live home by any arm.

## Notes

The branch is a single commit cut from `adolfousier/main`, carrying only this fix; `Issue-Ref` names exactly one issue. The buggy string test is still present upstream, so this is a fix rather than a refactor of already-correct code.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/350


adolfousier (commented):
This is my favourite kind of fix: the bug and the bypass turn out to be the same defect seen from two sides, and naming that is most of the work.

The old guard tested a raw string prefix, which got both directions wrong. `profiles/family/USER.md` from the default home is a real sibling target and was refused, while `./profiles/ops/TOOLS.md` from a named-profile home genuinely does duplicate and slipped straight through. Replacing one string test with two predicates, one on the home's shape and one on the path's components, makes the refusal fire exactly where duplication actually occurs. `home_is_named_profile` reading `parent().file_name() == "profiles"` is the right test for `<base>/profiles/<name>`, and skipping `Component::CurDir` is what closes the bypass rather than papering over it.

Passing the home in as a parameter is what makes the whole thing testable against both shapes instead of whichever one the process happens to be running under, and the error message naming the resolved home is a real improvement over asserting the path is malformed when it is not.

Verified independently before saying any of that: `validate_opencrabs_path` has no callers outside the tool and its test file, so widening the signature has no blast radius; `opencrabs_home()` returns exactly the two shapes the predicate distinguishes; and the traversal, absolute, tilde and null-byte refusals are all still exercised against both home shapes rather than only the named one.

The tests are the right kind. They drive the predicates directly, they cover the accept case that regressed and the bypass case that was accepted, and they live in `src/tests/` with no inline module.

Two cosmetic things we are tidying on top, neither worth a respin:

- the test file imports `write_opencrabs_file::*` and then `write_opencrabs_file::validate_opencrabs_path` again, which the glob already brought in;
- inside `test_profiles_prefix_duplication_rejected` the comment says "Also reject just `profiles/` with nothing after" but the assertion below it is `profiles/ops/MEMORY.md`. The bare `profiles` case is genuinely covered, just over in the predicate test.

Nothing needed from you. Merging as is with a small cleanup commit on top.

PR #1637, opened 2026-09-19
fix(logging): install a crash-signal diagnostic handler for SIGBUS/SIGSEGV

leshchenko1979:
## What

A fatal fault signal (`SIGSEGV`, `SIGBUS`, `SIGILL`, `SIGFPE`, `SIGABRT`) is delivered to the
kernel's default disposition, so the daemon dies silently — recording neither the signal nor the
faulting address. A production instance was killed by `SIGBUS` under host memory pressure and left
no panic, no backtrace and no log line behind. This installs a crash-signal diagnostic handler that
addresses that blind spot.

## Why

Diagnosing a silent death from the journal alone is not possible: the kernel OOM-killer was
separately observed reaping children inside the unit cgroup rather than the daemon itself, so the
faulting signal is not attributable from the journal. The handler makes the death self-describing.

## Implementation

- **`src/logging/crash.rs`** (new, 687 lines) — two separable halves:
  - a **pure formatter** (`CrashFacts`, `render_header`, `render_backtrace`) that turns the facts
    into the record text with no I/O and no allocation, so it is unit-testable without forking a
    process;
  - the **async-signal-safe handler** itself, installed on a dedicated 64 KiB alternate signal
    stack (`sigaltstack` + `SA_ONSTACK`), using `SA_SIGINFO`, `SYS_gettid` on Linux for the tid,
    raw non-allocating direct writes to both stderr and a crash log, and a **re-raise of the true
    signal** at the end so the process still dies by the original signal and the exit status stays
    honest.
  - `FAULT_SIGNALS` covers exactly five signals: `SIGSEGV`, `SIGBUS`, `SIGILL`, `SIGFPE`, `SIGABRT`.
  - The crash-log path is resolved and leaked as a `CString` at install time, because the handler
    may not allocate.
  - x86_64 ABI register extraction, with a non-x86_64 stub.
- **`src/logging/mod.rs`** — re-exports `install_crash_handler` on `unix`.
- **`src/main.rs`** — installs the handler at startup, with a non-fatal failure path so a handler
  that cannot be installed never prevents the daemon from starting.
- **`src/tests/crash_signal_test.rs`** (new, 344 lines) — 11 unit tests over the formatter, plus a
  re-exec integration test that drives a child into a real fault and asserts that it records the
  crash *and* dies by the true signal.
- **`src/tests/mod.rs`** — registers the test module.

### One formatting detail worth calling out

The record is truncated as a **true prefix**, not a fragment. An earlier revision used an
all-or-nothing `push_bytes`, which could splice a partial multi-byte escape sequence into the
record. The final formatter truncates at a boundary that cannot split an escape sequence, checked
across every cap value.

## Verification

**CI gate (full, no `--fast`)** — job `PR-lane gates (bc2c2566736e0ee02438661be43feb93135822de)`:

https://github.com/leshchenko1979/opencrabs/actions/runs/35431305725

**Smoke** — 4-leg PASS against the running daemon:

- **L1 LINEAGE** — `git merge-base --is-ancestor 84b93ebab89f4512479c38e9975cf6cc43b17f6e origin/main`
  rc=0; the handler was shipped and swapped, and the live daemon was driven directly.
- **L2 IDENTITY** — the harvested commit is content-identical to the smoked artifact for **every
  file the feature touches**. `git rev-parse <ref>:<path>` blob comparison, harvest commit
  `bc2c2566736e0ee02438661be43feb93135822de` vs the smoked fork-main tip
  `84b93ebab89f4512479c38e9975cf6cc43b17f6e`:

  | path | blob | result |
  |---|---|---|
  | `src/logging/crash.rs` | `0301dd0d61229d0fd66cd6c6a0fa3e3dd8088837` | identical |
  | `src/logging/mod.rs` | `8ab84b88af0b1a98c39843c5c60dcfb7e5d78207` | identical |
  | `src/main.rs` | `fea5ca33960f0be41f81163ef40741faba441a49` | identical |
  | `src/tests/crash_signal_test.rs` | `860f6721ec1714ddc3e4d610e86ff8f859b9bd4e` | identical |
  | `src/tests/mod.rs` | harvest `d3172f1ff223aa330b2798aef33a362f3ba66a3a` | differs only by fork-only sibling module registrations |

  For the last row the difference is confined to unrelated lines: the harvested file omits a number
  of fork-only test-module registrations that do not exist upstream, and the feature's own line
  `pub mod crash_signal_test;` is present in both trees. No line of this change differs between the
  two.

- **L3 CI GATE** — the run above (full gate: formatting, clippy, tests).
- **L4 BEHAVIORAL** — a live probe against the running daemon observed `SigCgt=0x1000104ea`, i.e.
  all five fault signals are caught by the installed handler. Because the handler's bytes are
  identical between the smoked artifact and this commit, the behavioural leg transfers: the binary
  that was driven executes exactly these bytes.

## Notes

The change is additive and self-contained: one new module, one new test module, and two one-line
registrations. Nothing existing is modified in behaviour — when no fault occurs the handler is
never entered.

Original issue: https://github.com/leshchenko1979/opencrabs/issues/352


adolfousier (commented):
A daemon that dies without saying how is the worst kind of outage, and this closes that. The structural choice that makes the PR reviewable is splitting the pure formatter from the handler: `CrashFacts` plus `render_header` and `render_backtrace` allocate nothing and contain no unsafe, so 11 of the tests are ordinary unit tests rather than process-forking rituals, and only the last one has to drive a real fault.

Details worth naming because they are the ones people get wrong:

- decoding `si_code` per signal rather than through one table. `1` is `BUS_ADRALN` under SIGBUS and `SEGV_MAPERR` under SIGSEGV, and a signal-blind decoder would have mislabelled every record. For the SIGBUS case that motivated this, `BUS_ADRERR` versus `BUS_MCEERR_*` is the entire difference between memory pressure and failing hardware.
- re-raising the true signal with `SIG_DFL` restored and the signal unblocked, so the exit status stays honest and a supervisor still sees the real cause.
- resolving and leaking the crash-log path as a `CString` at install time, because the handler cannot allocate.
- the truncation being a true prefix, so a partial escape sequence can never be spliced into the record.
- the non-x86_64 stub returning an error rather than succeeding silently. Reading `ucontext_t` registers on the wrong architecture would report a garbage fault address as fact, and a visible absence beats a confident lie.

Checked the one thing that could have broken the build: `libc::backtrace` is a glibc extension and does not exist on musl. This crate's release workflow builds gnu, Windows and Darwin only, and the handler is gated to `linux + x86_64`, so no target regresses. Also confirmed `panic = "unwind"`, so catching SIGABRT does not intercept ordinary panics.

Three things we are folding in on top.

**1. Every non-x86_64 boot now logs a warning.** The stub always returns `ErrorKind::Unsupported` and `main.rs:44` logs it at `warn!`. On macOS and aarch64 Linux, which are three of the five release targets, that prints a warning at every single startup for a platform where the handler was never implemented. A warning should mean something went wrong, not that a feature does not exist here. We are matching on `Unsupported` and logging that at `debug!`, leaving `warn!` for a genuine install failure on a platform that should have succeeded.

**2. The alternate stack claim is true for one thread.** `crash.rs:610` says the dedicated stack means "a stack-overflow SIGSEGV is still diagnosable", but `sigaltstack` is a per-thread attribute and threads do not inherit it, so only the thread that ran `install_crash_handler` has one. Every tokio worker takes the fault on its ordinary stack, which is the case where a stack-overflow SIGSEGV cannot be handled. For the SIGBUS that motivated this it makes no difference, and installing per-thread would need a runtime thread hook that is out of scope here. We are correcting the comment to say what it actually guarantees, rather than shipping a claim that reads stronger than the mechanism.

**3. `backtrace()` is the one call in the handler that is not async-signal-safe.** glibc resolves the unwinder lazily on first use, so the first call can `dlopen` and allocate, inside a handler that is documented as allocation-free. The standard mitigation is cheap and we are applying it: call `backtrace` once at install, in ordinary thread context, so the lazy load is already done and the signal-time call is a pure stack walk.

Also noted, not changed: `crash.log` is opened `O_APPEND` and `is_log_file` only matches the `opencrabs.` prefix, so `cleanup_old_logs` never touches it. Only crashes write to it, so unbounded is academic.

Nothing needed from you. Merging with those three stacked on top.

Issue #1648, opened 2026-09-20
feat(decisions): tiered decision cache so classification decisions are paid for once (shadow mode first)


adolfousier:
## Problem

OpenCrabs agents make hundreds of small classification-shaped decisions (triage, routing, voice-gate checks, socialcrabs draft scoring, self-audit verdicts) and pay full model price every time, even when the input is an exact repeat of something already decided. The qmd-replacement Store rewrite explicitly dropped `llm_cache` (`src/memory/db.rs:9`), so the exact-hit tier does not exist at all.

Context: TypeSafe's Jev release (2026-09-16) popularized the "System One" decision-layer pattern, but the decisive lever in the community playbooks is not the model swap, it is the ladder: deterministic bypass, then **exact decision reuse via normalized-hash cache** (measured 96.5% hit rates on template-shaped inputs), then cheap model, then frontier. Full research notes: `~/.opencrabs/research/jev-decision-pyramid.md`.

## Proposal: tiered decision cache ("never pay twice for the same thought")

- `decision_cache` table in opencrabs.db: `key TEXT PK, tier_id, result_json, p NULL, margin NULL, policy_version, created_at, hits, last_used_at`.
- `key = sha256(tier_id + policy_version + normalizer_version + canonical_json)`, canonical_json masks volatile parts (timestamps, uuids, ips) the way the pagerduty playbook normalizes `[IP]/[UUID]/[PATH]`.
- Read path: exact cache hit -> optional L2 similarity (opt-in per tier, trust-gated, adversarial input escalates instead of reusing) -> live model. Write path only when `margin = |p - threshold|` clears a floor: borderline calls are not frozen.
- `policy_version` bump invalidates wholesale (rules/categories changed).
- One tool: `decide_cached(tier, input, ask) -> { result, cached, tier_answered }`.
- Telemetry: per-tier answered distribution + avoided-call count folded into `/usage` and mission-control report. This is the differentiator claim: the cache is commodity, the integrated pyramid plus cost-avoided accounting is not.
- Land in shadow mode first: log would-be hits, zero behavior change, measure real hit rates before trusting any of it.

## Non-goals

- No Jev/TypeSafe provider in this issue (its `/v1/systemone` API is not OpenAI-compatible; it is a separate provider-impl decision the pyramid does not depend on).
- No similarity-tier defaults: opt-in per tier_id only.

## Acceptance criteria

- Migration for `decision_cache` applied on existing DBs without data loss.
[...]

PR #1660, opened 2026-09-21
feat(decisions): decide_cached tool, [decisions] config, shadow mode (Refs #1648)

adolfousier:
## PR2 of #1648 (stacked on #1659 — base is `feat/decision-cache-storage`, retarget to main once that merges)

The `decide_cached` agent tool + `[decisions]` per-tier config + the shadow-mode evaluation instrument. Kill rule honored: shadow changes nothing the agent experiences, `mode = "off"` takes zero cache code paths, empty config = today's behavior exactly.

### What ships
- **`decide_cached(tier, input, ask)`**: keys `input` through PR1's normalize ring; `live` may reuse a fresh row without a model call, `shadow` (default) always asks the model but records the would-hit, `off` is a plain model call touching no DB.
- **`[decisions.tiers.<name>]`**: `policy_version` **required** (named load error), `mode` shadow|live|off, `ttl_hours` freshness gate, `margin_floor` write gate enforced by the PR1 repository (unskippable), `similarity` rejected at load (L2 unimplemented — no operator can believe a fuzzy ring is on).
- **`decision_stats` migration + repository**: per-tier `calls`/`would_hit`/`live_hit` persisted via single atomic upsert (concurrent sessions can't lose counts). This table is the release-day keep-or-cut evidence; if it's empty on release day, the feature is removed, not extended (Adolfo directive 2026-09-21).
- Seed template: commented `[decisions]` block in `config.toml.example`; tool registered as extended (`tool_search` discoverable), not core.

### Behavior identity proof
`shadow_always_calls_model_and_counts_would_hit`: a second identical ask under shadow re-calls the stub provider (assertion on call count) and flips `would_hit` true with counters persisted. `live_first_miss_then_cached_hit` proves `cached=true` happens only under `mode=live`. `contract_violation_is_not_cached` + `sub_floor_margin_is_answered_but_not_stored` pin the two write gates.

### Migration note
`20260921000002_add_decision_stats.sql`: idempotent CREATE, appended last per the list invariant, removal is `DROP TABLE IF EXISTS decision_stats`.

Refs #1648 (PR2 of 4; #1659 is PR1).


adolfousier:
Shipped to main: head c3e1f16a3 is an ancestor of 1f0393414 (origin/main). Stacked PRs never auto-closed because their base is a sibling feature branch, not main, and GitHub's merge detection fires on the base branch. Closing as done; CI was green on this exact tip (9008 passed / 0 failed).

PR #1662, opened 2026-09-21
feat(decisions): /usage accounting block, mission-control counters, TTL sweep (#1648) PR3/4

adolfousier:
Refs #1648 - part 3/4 of the decision-pyramid stack (stacked on #1660, which is stacked on #1659).

## What
The measurement half of L1: the shadow counters PR2 persists now have a surface, and the cache has a pruning path.

- **/usage decisions block** (both renderers: channel format_usage and the slash tool): per tier, mode + policy version, calls, would-hit %, live-hit %, rows cached, and an estimated calls-avoided line. Renders **iff** decision_stats has counters; silent otherwise. Silence is a feature: an unmeasured tier cannot fake a dashboard, and an empty block on release day is itself the cut signal.
- **Mission Control report**: Decision cache table from the same counters (point-in-time cumulative, deliberately unwindowed like the RSI counters).
- **TTL pruning at startup**: per-tier ttl_hours deletes rows whose COALESCE(last_used_at, created_at) is stale (recently-reused rows survive); orphan rows for removed tiers are swept; an empty known-tier list clears the whole table, which is the one-line kill-switch removal.
- New repo queries: count_by_tier, prune_expired, delete_unknown_tiers; pure renderer in src/decisions/report.rs.

## Tests
src/tests/decisions_report_test.rs: the sentinel (renders iff counters exist), TTL prune keeps a refreshed row and drops a stale one, orphans swept / known kept, empty-known clears, counters seeded from bumps render end-to-end, default mode is shadow. Mission-control report tests updated (section present with counters, omitted on defaults).

12 files, +572/-6, commit 0d4ae8b3c.

Release-day evaluation vs the >=30% would-hit / >=100 calls bar still owed in PR4.

Refs #1648 (do not close; PR4 carries the evaluation + docs and closes it).


adolfousier:
Shipped to main: head 0a4d50724 is an ancestor of 1f0393414 (origin/main). Same stacked-base reason for no auto-close. Closing as done; CI green on 9015 passed / 0 failed at final stack tip 8f3ee3188.

Issue #1673, opened 2026-09-21
feat: ACP client-parity gaps (replay, catalog, modes, compact, commands, usage size, quiet)


moneyacademyKE:
## Summary

Follow-up to #1540 (merged, shipping in v0.5.2): the ACP server speaks the protocol, but GUI clients (MonoCode, Zed) still see a bare-bones agent. This tracks the parity layer that makes the merged server fully usable as a harness peer — each item verified end-to-end against a live `opencrabs acp` process on a fork branch.

## Scope (one layer: client-experience parity)

- **Live model catalog** — `session/new` advertises `models.availableModels` from the provider registry (`provider/model` pair ids); `set_model` becomes pair-aware (swaps the session's provider). Clients get a real model picker instead of "default".
- **Native `session/set_mode`** — real mode vocabulary (`plan`, `auto-accept-edits`, `auto`, `full-access`, `supervised`), enforced server-side in the approval callback: plan denies mutations so the agent learns the boundary from the denial. Extends the merged alias behavior.
- **`session/compact`** — drives the loop's existing manual-compaction marker turn; claims the in-flight slot exactly like `session/prompt` so a compact cannot race a live turn.
- **Command discovery** — `available_commands_update` push at session creation (built-ins + skills + commands.toml). ACP-native discovery; no filesystem scraping.
- **Transcript replay on `session/load`** — stored history re-sent as `session/update` notifications before the load answers; `loadSession` flag flips to `true` to match. Per-session model restored from the session row.
- **Usage `size`** — usage updates carry the context-window total alongside `used`.
- **`--quiet` for `opencrabs run`** — machine-pure stdout for one-shot callers (title/git-text generation parses a single JSON object).

## Non-goals

Cost reporting, `warmupText`, structured questions — deliberately out.

## Verification

Built and smoke-tested on `moneyacademyKE/opencrabs@feat/acp-parity2` (rebased onto current main after the #1540 merge): `cargo fmt`/`clippy` clean, ACP test suite green, four-phase NDJSON smoke (handshake, cross-process resume, replay-before-load, model persistence across processes, streamed `end_turn`).

PR #1674, opened 2026-09-21
feat(acp): client parity — replay, catalog, modes, compact, commands, usage size, quiet run

moneyacademyKE:
Implements #1673 — the client-experience parity layer for the ACP server mode merged in #1540 (v0.5.2).

Each commit lands one feature, composed on the merged base rather than reworking it — the maintainer's on-top fixes (permission timeout, TOCTOU-safe in-flight slot, test placement) are kept and extended, not reverted:

- `f728275e` catalog — `models.availableModels` in `session/new` from the provider registry; pair-aware `set_model` (keeps the `modeId` alias)
- `6c8f181d` set_mode — real mode vocabulary, server-side approval policy: `plan` denies mutations outright; the approval callback takes mode + cancel together (timeout preserved)
- `d59697ae` compact + commands — `session/compact` drives the manual-compaction marker turn, claims the in-flight slot like `session/prompt`; `available_commands_update` push at session creation
- `39c71888` `run --quiet` — machine-pure stdout for one-shot callers
- `0617a259` replay + model restore — `session/load` replays stored history as updates before answering; `loadSession: true` (with the honesty test updated to the new contract); per-session model restored from the session row
- `ee90bc3c` usage `size` — context-window total alongside `used`
- `0559d37e` steer ext-prefix — `_session/steer` with `session/steer` kept as alias
- `88b82e8e` set_model persistence — the session-row write that makes model restore real
- `83630a8` replay completeness — covers upstream's `Segment::Blocked` variant (phantom-blocked narration replays as a labeled thought chunk, mirroring the TUI's rendering); replay fixtures aligned with the parser's line-start marker contract

Verification: `cargo fmt --check` clean, `clippy` clean, ACP tests green, four-phase NDJSON smoke against a live process (handshake → cross-process resume → replay-before-load → model persistence → streamed `end_turn`).

The upstream tests the maintainer added are updated where the feature changes the contract (`initialize_result_is_honest_about_load_replay` now asserts the flag matches the new replay behavior, not the old static value).


moneyacademyKE:
Follow-up fix on this branch: the doubled streamed answer reported in the PR body's footnote is diagnosed and fixed.

**Wire evidence:** the four-phase smoke captured two identical `agent_message_chunk` frames ("ACP-OK" twice) for a single one-word answer.

**Root cause:** the agent loop emits live deltas via `StreamingChunk` and then the round's full text again via `IntermediateText`; the ACP bridge mapped both to `agent_message_chunk`, so every streamed answer rendered twice on the client.

**Fix (540acd0):** track the text streamed since the last round boundary and skip the aggregate when it repeats it — the same dedup rule the WhatsApp channel applies pre-send. Aggregates carrying unstreamed text (CLI providers stream nothing) still emit, and tool events reset the boundary so a round-2 answer dedups against its own chunks. Whitespace is trimmed on both sides; an empty aggregate is never treated as a duplicate.

**Coverage:** 3 new unit tests on the dedup predicate (exact repeat, whitespace disagreement, unstreamed/differing text), and the smoke now *asserts* the joined answer text equals `ACP-OK` exactly — the old assertion tolerated the doubling, so this class of regression can no longer pass silently.

**Live proof:** the four-phase smoke re-ran on the rebuilt binary: `answer undoubled? true` — the joined answer text is exactly `ACP-OK`.

---

Second fix on this branch, found by clippy of all things: `session/set_mode` was routed to `set_model` — an over-eager conflict resolution during the rebase onto merged main folded `SESSION_SET_MODE` into the `SESSION_SET_MODEL` arm, leaving the real mode handler unreachable. The visible symptom was one `unreachable pattern` warning; the behavioral symptom was that every `set_mode` request was answered as a bogus model switch (the mode id read as a model name). Clippy's dead arm was the only clue — the unit tests stayed green because they test the handlers, not the routing.

Fixed by restoring the two dispatch arms, and `set_model` no longer accepts a `modeId` param — a mode id reaching the model handler is a routing accident and now fails loud instead of silently switching models. The smoke gained a routing pin: a valid `set_mode` must return `{}`, a bogus modeId must return `-32602`.

moneyacademyKE:
Follow-up on this branch: **the context meter now restores on `session/load`.**

**The gap:** replay restored the transcript but never a usage event — so after restarting MonoCode, the context meter rendered nothing until the first new turn completed, exactly when a restored long session most needs pressure visibility.

**The fix:** `session/load` emits one usage update after the replay. The `used` value is the provider's *own* last measurement — `Message.input_tokens`, whose column doc already calls it "the authoritative last known context size on session load" — falling back to the session row's lifetime total when the session predates usage persistence. `size` comes from `context_limit_for_session`. No tokenized estimates of raw content anywhere: the meter shows a real number or nothing.

Two new protocol tests pin the selection rule (last assistant row, not first, not a sum, not a user row) and the pre-persistence fallback shape. 32/32 ACP tests green.

The four-phase smoke now asserts the usage restore lands as a `session/update` before the load response.

Live proof from the smoke run just now (`443dc35e`): `usage restored? 41818/1000000` — the provider's actual measurement, restored on a fresh process, alongside the full four-phase pass (handshake / resume / `-32601` / streamed `end_turn` with the undoubled answer / replay-before-load / model persistence).

moneyacademyKE:
**fix: per-call approval override beats config auto-approve policy** (227ba803)

**Root cause found by dogfooding the client end-to-end:** with `approval_policy = "auto-always"` in config, an ACP session declaring `set_mode supervised` never asked the client — no permission dialog, ever. The loop's per-call override callback was passed, mode-aware, and correct; but the tool-execution context seeded `auto_approve` from the config policy, and the gate's third condition (`!tool_context.auto_approve`) silently defeated the override precedence the loop's own comment promises.

**Fix (one line):** `context_auto_approve = auto_approve_tools && !has_override_approval`. Post-approval flips are untouched — tools still see approval once granted. Surfaces that pass no override (TUI, CLI, subagents) are byte-for-byte unchanged.

**Test:** `test_override_callback_beats_auto_approve_policy` pins the exact dogfood setup (auto-always config + per-call override must consult the callback). Suite: 8/8.

**Live proof (real `opencrabs acp` child, real provider):** prompt → agent calls bash → `session/request_permission` on the wire → client renders Allow/Deny → Allow → turn completes with real tool output. This path was dead for any user running auto-approve locally.

Issue #1691, opened 2026-09-22
feat(config): warn at load when timeout_secs sits below stream_idle_timeout_secs (#1687)


adolfousier:
Part of #1687.

## Why the #1687 defect stayed invisible for so long

The incident config set:

```toml
[providers.zai]
timeout_secs = 90                 # non-streaming ceiling (and, pre-#1687, the stream killer)
stream_idle_timeout_secs = 300    # inter-chunk silence tolerance
```

The idle tolerance (300s) is **larger** than the total ceiling (90s), which is structurally
incoherent: a stream can never be allowed 300s of silence inside a 90s lifetime. The transport
ceiling always wins, on every turn, and it wins on a healthy connection. Nothing warned.

The same inversion is what made #1635 read like a fixed number problem: raising 60s to 300s moved
the wall instead of removing it, and the guard added for it
(`src/tests/provider_stream_total_timeout_test.rs:46-55`) only checks that the total stays at least
5x the *hardcoded* 20s idle constant. It cannot see a user-supplied inversion.

## Acceptance

- At config load, for every provider section (and the `[agent]` level once #1687's sibling issue
  lands), warn when a resolved `timeout_secs` is below the resolved `stream_idle_timeout_secs`,
  naming the provider, both values, and which one will fire first.
- Warning only, no rejection: a user may legitimately want a tight non-streaming ceiling, and after
  #1687 the two clocks no longer compete on the stream path.
- A test per branch: inverted (warns), equal (silent), inverted-but-zero/disabled (silent), and
  provider-overrides-agent resolving to a non-inverted pair (silent).

adolfousier:
Withdrawn before implementation.

Two reasons, both from correcting the parent's evidence:

1. **Its premise was a config value that does not exist.** The motivating example here was
   `timeout_secs = 90` with `stream_idle_timeout_secs = 300`. The reporting host's
   `[providers.zai]` carries neither key, and `factory.rs:1997` (`if let Some(secs) =
   config.timeout_secs.filter(|&s| s > 0)`) only applies the flag when it is present. The clocks
   in play were our own hardcoded defaults, not an operator's disagreement between two knobs.
2. **Once #1687 lands there is nothing left to warn about.** The whole point of the parent fix is
   that `timeout_secs` stops governing the stream path — streams move to a client built without
   `.timeout()`. A non-streaming ceiling can no longer starve an inter-chunk guard, because they
   bound different things on different clients. Any load-time check for `timeout_secs <
[...]

Issue #1690, opened 2026-09-22
fix(agent): thinking_loop_timeout_secs kills a delivering stream instead of nudging, and cannot be scoped per provider (#1687)


adolfousier:
Part of #1687.

## Intended semantics

`thinking_loop_timeout_secs` is the "nothing is happening" clock: the model has produced no tool
calls for this long, so nudge it to act. It is a prompt to the model, not a verdict on the
connection, and it must not stop work that is genuinely in flight.

## What the code does

`src/brain/agent/service/helpers.rs:485-537` races the stream against a deadline and, when it fires
with `!has_tool_call` (`:529`), does:

```rust
tracing::warn!("🧠 Thinking-loop timeout (#890): {}s elapsed with zero tool calls. \
                Killing stream for phantom enforcement retry.", ...);
return Err(crate::brain::provider::ProviderError::ThinkingLoopTimeout(
    thinking_loop_timeout_secs,
));
```

So a healthy, still-delivering stream is torn down at 600s because the answer it is writing happens
to need no tools. `tool_loop.rs:2004-2014` handles that error by retrying with phantom enforcement,
which re-sends the whole conversation. Two mismatches:

| Mismatch | Evidence |
|---|---|
| Kills the stream instead of nudging | `helpers.rs:535` returns `Err(ThinkingLoopTimeout)`, gated on `if !has_tool_call` (`:529`), so a long prose-only answer trips it |
| No per-provider scoping | the value is read from `Config::current().agent.thinking_loop_timeout_secs` (`:491-497`) with no provider-level lookup, unlike `stream_idle_timeout()` which does consult the provider (`:477`) |

Interaction with #1687: wherever the transport total ceiling sat below this 600s guard, streams
died at the ceiling first and this guard was masked on long turns (the compat family shipped at 60s
from #217 until #1635 raised it to 300s). With #1687 landing, the ceiling leaves the stream path
entirely and this becomes one of the first clocks to actually fire on a long turn — which is why it
needs its own issue.

## Acceptance

- On expiry with no tool call, the turn is nudged (the existing phantom-enforcement retry path)
  without discarding a stream that is still delivering chunks; state the resulting contract in the
  test name.
- `thinking_loop_timeout_secs` resolvable per provider, `[agent]` as fallback, default 600
  (`config/types.rs:1581-1583`) unchanged, `0` still disables it, CLI providers still exempt
  (`helpers.rs:491`).
[...]

Issue #1689, opened 2026-09-22
fix(provider): anthropic and gemini silently drop timeout_secs and stream_idle_timeout_secs (#1687)


adolfousier:
Part of #1687.

## Symptom

`[providers.anthropic] timeout_secs = 120` in config.toml does nothing, with no warning. Same for
`stream_idle_timeout_secs`. The keys are accepted by the parser and then thrown away, so the user
believes they configured the provider.

## Root cause

Neither struct exposes a setter, and the factory never asks them to configure anything:

| Fact | Evidence |
|---|---|
| `AnthropicProvider` has fields `api_key`, `client`, `custom_default_model`, `configured_context_window` and no timeout member | `src/brain/provider/anthropic.rs:40-46` |
| Its client is built inline with the const, `.timeout(DEFAULT_TIMEOUT)` commented "Total request timeout (including streaming)" | `anthropic.rs:52-60`, const at `:29` |
| `GeminiProvider` identical | `gemini.rs:44-46`, `:91-99` |
| Both are constructed plain by the factory, never passed the `ProviderConfig` | `factory.rs:2246` (`AnthropicProvider::new(api_key)`), `factory.rs:2064` (`GeminiProvider::new(api_key).with_model(model)`) |
| The only reader of the config keys is the OpenAI-compatible configurator | `factory.rs:1997-2003` |
| The fields themselves exist | `config/types.rs:2841` (`timeout_secs`), `:2846` (`stream_idle_timeout_secs`) |

The compat family is not blame-free either: `with_stream_idle_timeout` (`custom_openai_compatible.rs:2496-2499`)
only stores the value, so the idle key has no transport effect at all, while the total key does.

## Related stale doc

`config/types.rs:2839` documents `timeout_secs` as "Overrides the default 60s client timeout". The
default has been 300s since #1635 (`custom_openai_compatible.rs:27`, asserted against the native
providers by `src/tests/provider_stream_total_timeout_test.rs:28-43`). Fix the line here.

## Acceptance

- `[providers.anthropic]` / `[providers.gemini]` `timeout_secs` and `stream_idle_timeout_secs` reach
  the provider, with a test per key per family asserting the resolved value on the built provider.
- An unrecognized-key or "ignored" warning is out of scope here; the value must simply land.
- Depends on #1687 for what `timeout_secs` is allowed to govern (non-streaming only).

adolfousier:
[...]

Issue #1688, opened 2026-09-22
feat(config): resolve timeout flags [providers.<name>] -> [agent] -> default across every provider family (#1687)


adolfousier:
Part of #1687.

## What exists today

`timeout_secs` and `stream_idle_timeout_secs` are read in exactly one place:
`configure_openai_compatible` (`src/brain/provider/factory.rs:1997-2003`), which applies them via
`with_timeout` / `with_stream_idle_timeout` on the OpenAI-compatible family only. There is no
`[agent]`-level timeout key at all: `AgentConfig` (`src/config/types.rs:1267`) carries
`thinking_loop_timeout_secs` and nothing else in this family.

So today the same key means three different things depending on which provider you write it under:

| Written under | Effect |
|---|---|
| `[providers.openai]`, `[providers.zai]`, `[providers.custom.*]`, qwen/xiaomi/dashscope/openrouter/ollama | applied (`factory.rs:1997-2003`) |
| `[providers.anthropic]`, `[providers.gemini]` | parsed into `ProviderConfig` and silently dropped (see the sibling issue) |
| `[agent]` | not a known key |

## Target state

One resolution for both keys, honoured by every provider family:

```
[providers.<name>]  →  [agent]  →  compiled default
```

- `timeout_secs` default **300s** for every non-local provider, openai-compatible included, and it
  is a **non-streaming** ceiling only (#1687 removes it from the stream path).
- Local providers keep their current behaviour (no total cut, 3600s inter-chunk tolerance) but the
  key must still be settable under their own `[providers.<name>]` section so a user can scope it.
- `stream_idle_timeout_secs` keeps its tier defaults (3600s CLI/local, 45s z.ai on `api.z.ai`,
  20s other remote, `helpers.rs:477-482`) and gains the same two-level resolution.

## Acceptance

- `[agent] timeout_secs` and `[agent] stream_idle_timeout_secs` deserialize and take effect for
  every family, with the per-provider value winning.
- A test per level: provider-only, agent-only, both set (provider wins), neither set (default).
- README `### Per-Provider Timeouts` (`README.md:1565`) renamed and rewritten to state the
  resolution order, so the key is no longer documented as per-provider-only.

adolfousier:
Still open, correctly: #1692 does not touch resolution. Verified on `ada6901e8`, `factory.rs:1997` and `:2001` read the per-provider struct and nothing else:

```rust
[...]

PR #1692, opened 2026-09-22
fix(provider): streams must not inherit the transport total wall clock (#1687)

### LINKED ISSUE #1687: Streams inherit the transport total wall clock, so every stream longer than timeout_secs dies and falls back to the chain (#1635 raised the number, never removed the ceiling)
## Symptom (reported)

GLM 5.3 flash on z.ai fell back to the fallback chain on **every single turn**, while the same
account and keys on other harnesses ran 2h+ tasks with no errors and no fallback. No rate limit
was ever hit: nothing about the provider's response justified a fallback.

## The config contributed nothing

`[providers.zai]` on the reporting machine carries no timeout keys at all:

```toml
[providers.zai]
vision_model = "glm-5.3-flash"
default_model = "glm-5.3"
enabled = true
endpoint_type = "coding"
models = [...]
```

And the factory only applies the flag when it is present:

`src/brain/provider/factory.rs:1997` — `if let Some(secs) = config.timeout_secs.filter(|&s| s > 0)`

So **no user setting was involved**. The wall clock that governed those streams is ours, hardcoded
in the provider family. That is the difference between "one operator misconfigured a knob" and
"the shipped default cuts every long stream for everybody" — it is the second one.

## Root cause

The reqwest **total request timeout** is applied to the client that streams the SSE body.
reqwest's `.timeout()` covers the response body read, so it is a hard wall-clock ceiling on a
stream no matter how healthily it is delivering chunks. Every stream that outlives the ceiling
dies mid-body, and the tool loop's layer-2 recovery re-sends the byte-identical request up to
`MAX_STREAM_RETRIES` times into the same wall before it ever asks the chain.

Evidence on `0b6cf46cb`, all three provider families:

| Family | Total timeout set | Default value when the flag is unset | `stream()` uses that client |
|---|---|---|---|
| OpenAI-compatible (z.ai, every `custom.*`, local) | `custom_openai_compatible.rs:37` `.timeout(timeout)` in `build_http_client` (`:35`), fed `DEFAULT_TIMEOUT` at `:28`; comment `:22-27` states it covers streaming | 300s | `:3616 fn stream()` → `self.client.post` at `:3707`, `:3751`, `:3782` |
| Anthropic | `anthropic.rs:53` `.timeout(DEFAULT_TIMEOUT)`, inline comment "Total request timeout (including streaming)" | 300s | `:315 fn stream()` → `.client.post` at `:335` |
| Gemini | `gemini.rs:93` `.timeout(DEFAULT_TIMEOUT)` | 300s | `:519 fn stream()` → `.client.post` at `:545` |

`complete()` and `stream()` share one client per family (`compat:3414` / `compat:3616`), so the
same value is simultaneously the non-streaming ceiling and the stream guillotine. `with_timeout`
(`compat:2489-2493`) rebuilds that shared client, so setting the flag shortens the guillotine too.

### The number has been moved, never removed

`git log -S` on the compat constant:

| Commit | Change |
|---|---|
| `857a04e6f` (#217) | 300s → **60s** |
| `cca5534ee` (#1635, 2026-09-19) | 60s → 300s, plus `src/tests/provider_stream_total_timeout_test.rs:5-14` documenting the total as an SSE ceiling |

#1635 hit this exact symptom and "fixed" it by raising the number, which left the mechanism
intact. Any deployment older than 2026-09-19 was cutting compat streams at **60s**.

### What is verified and what is not

Verified: the mechanism fires, deterministically, with no config involved. The regression test in
this PR stands up a local SSE server delivering a fragment every 120 ms against a 250 ms total
timeout and the stream dies with `Streaming error: error decoding response body` — the same
undifferentiated text #1686 describes, which is why this never read as a timer.

Not verified: which clock fired on the reporting host's turns. Those logs are not in this
session. If the host ran a pre-#1635 binary the answer is the 60s default; either way the fix is
the same, because the ceiling should not be on the stream path at all.

## Fix in scope (this issue)

Streams must not inherit the transport total wall clock. Split the client per family:

- a **stream client** built with `connect_timeout` (10s, still bounds a dead handshake), pool and
  `tcp_keepalive` settings, and **no** `.timeout()`;
- the existing client keeps `.timeout()` and serves `complete()`, `/models` and every
  non-streaming path;
- `stream()` and its in-function retry sites move to the stream client.

Inter-chunk silence remains the app-level guard in `helpers.rs:477-482` (provider value, else
3600s for CLI/local, else 20s remote; z.ai gets a host-aware 45s via `factory.rs:1784-1800` →
`zhipu_endpoint.rs:62-69`), which is the correct detector for a genuinely dead stream. No new
clock is introduced, so no new disagreement.

## Explicitly out of scope (child issues)

1. #1688 — unify the timeout flags: resolution `[providers.<name>]` → `[agent]` → default across
   every family, one documented default instead of per-family constants that drifted (120s → 300s
   → 60s → 300s in the compat family alone).
2. #1689 — Anthropic and Gemini silently drop `timeout_secs` / `stream_idle_timeout_secs`: neither
   struct has a setter and `factory.rs` builds them plain, so the keys are parsed into
   `ProviderConfig` (`config/types.rs:2841`, `:2846`) and discarded. Only
   `configure_openai_compatible` reads them.
3. #1690 — `thinking_loop_timeout_secs` (default 600, `config/types.rs:1581`) **kills** the stream
   (`helpers.rs:535` returns `ProviderError::ThinkingLoopTimeout`) instead of nudging the model to
   emit tool calls, and cannot be scoped per provider.

Withdrawn: the "warn when `timeout_secs < stream_idle_timeout_secs`" child. It was filed against a
config value that does not exist, and once this issue lands the two clocks cannot disagree at all
— `timeout_secs` stops governing streams, so it can never starve the idle guard.

adolfousier:
## What this changes

A stream no longer inherits the transport **total** wall clock, in all three
provider families.

reqwest's `.timeout()` covers the response body read. The client that carried it was shared by
`complete()` and `stream()`, so the same value was simultaneously the non-streaming ceiling and a
hard cut on every SSE body — no matter how healthily chunks were still arriving.

## Why this is not about a user setting

`[providers.zai]` on the reporting host carries **no** timeout keys, and `factory.rs:1997` applies
the flag only when it is present and > 0:

```rust
if let Some(secs) = config.timeout_secs.filter(|&s| s > 0) {
    provider = provider.with_timeout(std::time::Duration::from_secs(secs));
}
```

So the clock that governed those GLM 5.3 flash turns was ours, hardcoded in the family default —
300s today, and **60s from #217 until #1635 raised it on 2026-09-19**. Any deployment older than
that was cutting every long compat stream at 60s. That is the difference between one operator
misconfiguring a knob and the shipped default doing this to everybody.

| Family | Total timeout set | `stream()` used that client |
|---|---|---|
| OpenAI-compatible (z.ai, every `custom.*`, local) | `custom_openai_compatible.rs:37`, fed `DEFAULT_TIMEOUT` (`:28`) | `:3616 fn stream()` → `self.client.post` at `:3707`, `:3751`, `:3782` |
| Anthropic | `anthropic.rs:53` (inline comment said "including streaming") | `:315 fn stream()` → `.client.post` at `:335` |
| Gemini | `gemini.rs:93` | `:519 fn stream()` → `.client.post` at `:545` |

## The fix

Per family: a **stream client** with `connect_timeout` (10s, still bounds a handshake that never
lands), pool settings and `tcp_keepalive`, and **no** `.timeout()`; the existing timed client keeps
serving `complete()`, `/models` and every buffered path. `stream()` and its in-function retry sites
move to the stream client. `with_timeout` now rebuilds only the non-streaming client, and
`AnthropicProvider::with_client` cannot reintroduce a ceiling on the stream path — an injected
client governs non-streaming calls, the stream client is always the total-timeout-free one.

`stream_idle_timeout_secs` remains the only timer a stream meets (3600s CLI/local, 45s z.ai on
`api.z.ai`, 20s other remote), which is the correct detector for a genuinely dead stream. No new
clock is introduced, so no new disagreement between clocks.

## Proof it fires, and proof the fix holds

`src/tests/stream_total_wall_clock_test.rs` — a local chunked-SSE server delivering one fragment
every 120ms (~1.4s total) against a deliberately tiny 250ms total ceiling:

- **Pre-fix, on `0b6cf46cb`:** `stream was cut: Streaming error: error decoding response body`.
  That is the #1687 defect reproduced, and the undifferentiated error text is why it never read as
  a timer — see #1686.
- **Post-fix:** all 12 fragments delivered, test passes.

The same file keeps the ceiling honest where it belongs: a 900ms buffered body against a 150ms
ceiling still fails (with retries disabled, so the assertion measures the clock and not
`complete()`'s 4-attempt ~15s retry budget — the first run of that test reported 14.86s for a 150ms
ceiling for exactly that reason), and a control proves a response inside the ceiling still succeeds.

`src/tests/provider_stream_total_timeout_test.rs` — #1635's guard compared the total against the
idle window. That relationship no longer exists (different clients, different clocks), and the
premise was wrong anyway: the defect was never the **value** of the number, since #1635 raised it
60s → 300s and the mechanism survived untouched. Replaced with a source scan of all three families:
the stream builders carry no `.timeout(`, and `stream()` posts on `stream_client`. #1635's
cross-family floor and the retry-log-budget scan are kept.

## Docs corrected

`README.md` (feature row + Per-Provider Timeouts) and `src/config/types.rs` said `timeout_secs`
"caps a whole request" / "Overrides the default 60s client timeout". Both were wrong twice over:
the default has been 300s since #1635, and the key never should have touched streams. Now stated as
the non-streaming ceiling it is, with the stream path called out explicitly.

## Out of scope (children of #1687)

- #1688 — unify flag resolution `[providers.<name>]` → `[agent]` → default across every family,
  instead of per-family constants that drifted 120s → 300s → 60s → 300s in the compat family alone.
- #1689 — Anthropic and Gemini silently drop `timeout_secs` / `stream_idle_timeout_secs` (no
  setters on either struct; `factory.rs` builds them plain, so the keys parse and are discarded).
- #1690 — `thinking_loop_timeout_secs` kills a delivering stream instead of nudging, and cannot be
  scoped per provider. With this PR it becomes one of the first clocks a long turn actually meets.
- #1691 — **withdrawn** (closed as not planned): filed against a config value that does not exist,
  and once this lands the two clocks cannot disagree at all.

## Verification

`arch -arm64`, on the exact commit, whole tree including tests:

| Gate | Result |
|---|---|
| `cargo fmt --all` | exit 0, no reformat needed |
| `cargo clippy -p opencrabs --all-features --tests -- -D warnings` | exit 0 |
| `cargo test --lib -- stream_total_wall_clock provider_stream_total_timeout docs_provider_timeout` | `test result: ok. 8 passed; 0 failed; 9075 filtered out` |

Full-suite receipt is CI's on this PR (this box is a Mac and does not carry the whole build here).

---
Closes #1687. Children #1688 #1689 #1690 stay open; #1691 withdrawn as filed on a config value that does not exist.


adolfousier:
Two commits on `fix/stream-no-total-wall-clock`, tip `ada6901e8`.

**Root cause (corrected from the original report).** Not a user setting. `[providers.zai]` on the reporting host carries no timeout keys at all, and `factory.rs:1997` applies the flag only when present and > 0:

```rust
if let Some(secs) = config.timeout_secs.filter(|&s| s > 0) {
    provider = provider.with_timeout(Duration::from_secs(secs));
}
```

So the clock that guillotined GLM 5.3 flash was ours, hardcoded per family: `DEFAULT_TIMEOUT` in `custom_openai_compatible.rs:28`, which was 60s from #217 until #1635 raised it to 300s on 2026-09-19. Any binary older than that cut every long compat stream at 60s with zero config involved.

**What changed.** `fc4095e99` splits a total-timeout-free stream client into all three families (`custom_openai_compatible.rs` +56/-15, `anthropic.rs` +32/-2, `gemini.rs` +23/-1). Each gets `connect_timeout` 10s, pool settings and `tcp_keepalive`, and no `.timeout()`; the timed client keeps serving `complete()`, `/models` and buffered paths. `with_timeout` now rebuilds only the non-streaming client, and `AnthropicProvider::with_client` cannot reintroduce a ceiling on the stream path. `stream_idle_timeout_secs` is now the only timer a stream meets.

**Tests.** `src/tests/stream_total_wall_clock_test.rs` (new, 3 tests) drives a local chunked SSE server at one fragment per 120ms against a 250ms total ceiling: pre-fix on `0b6cf46cb` it reproduces the defect (`stream was cut: error decoding response body`), post-fix all 12 fragments arrive. Test 2 pins `RetryConfig { max_attempts: 0 }` so the elapsed assertion measures the transport clock and not `complete()`'s 4-attempt backoff; the unpinned first run reported 14.855s against a 150ms ceiling for exactly that reason. `provider_stream_total_timeout_test.rs` replaces #1635's total-vs-idle const comparison with a source scan (stream builder exists, body carries no `.timeout(`, `stream()` posts on `stream_client`); that comparison guards nothing once the two clocks live on different clients.

**Receipts.**

| Gate | Scope | Result |
|---|---|---|
| `cargo fmt --all --check` | cleaned tree | exit 0 |
| clippy, exact CI shape (`--locked --lib --bins --tests --examples --all-features -- -D warnings`) | cleaned tree | exit 0 |
| targeted tests | cleaned tree | 8 passed, 0 failed, 1.76s |
| `cargo test --locked --profile ci --all-features --lib` | `fc4095e99` | 9058 passed, 0 failed, 34 ignored, 36.66s |

`ada6901e8` deletes 34 dead lines of a superseded test helper plus the `std::sync::{Arc, Mutex}` import only it needed. CI's Lint job failed on it (`function serve_hanging_json is never used`) and that failure skipped the hosted `Test` job entirely, so no hosted suite receipt exists for this branch yet. The deletion is inert, but the `Test` job on this tip is what closes that gap.

Not merged: Lint is pending on `ada6901e8` and merging to main is not mine to do.

adolfousier:
**Sha correction after branch rebase (2026-09-22 ~20:10 UTC).**

Earlier remarks on this PR cite `fc4095e99` and `ada6901e8`. Those shas are no longer the branch head. `main` moved to `a7372ee94` while the PR was in flight, and syncing it with `update-branch` created a merge commit at the head. The `main` ruleset 13432403 enforces `required_linear_history`, so that merge commit blocked the merge even with every required check green:

| Required context | State on `fb23441a2` (pre-rebase) |
|---|---|
| Lint | pass 5m6s |
| Test | pass 14m54s |
| CI Required Gate | pass 6s |
| `mergeStateStatus` | **BLOCKED** (base branch policy) |

Repair, in order:

1. Force-pushed the branch back to `ada6901e8`, dropping only the sync merge commit. Verified lossless first: that commit added exactly `a7372ee94`'s own two files (`compaction_prompts.rs`, `compaction_prompts_test.rs`), which `main` already carries. No PR content was lost.
2. Asked GitHub to restack server-side (`update-branch` with `update_method=rebase`), so no commit was ever rewritten locally.

New head, linear on current `main`:

| New sha | Was | Subject |
|---|---|---|
| `5bee1dc5c` | `fc4095e99` | `fix(provider): streams must not inherit the transport total wall clock (#1687)` |
| `3b4b26f98` | `ada6901e8` | `test(stream): drop the unreferenced hangup recorder` |

Diff verified byte-identical to the pre-rebase PR: 8 files, +426/-34, `src/tests/stream_total_wall_clock_test.rs` at 234 lines, zero `serve_hanging_json` hits.

CI is re-running on `3b4b26f98`. Repo-level auto-merge is disabled, so the merge is manual and lands only once `Lint`, `Test` and `CI Required Gate` are green on this head.

Issue #1249, opened 2026-08-27
[providers.fallback] chain never hot-reloads — a provider deleted from fallback_chain keeps serving live traffic until restart


adolfousier:
## Symptom

Remove a provider from `[providers.fallback].fallback_chain` in `config.toml`, save. The ConfigWatcher fires and reports a reload, but the removed provider **keeps serving turns**.

Observed live:

```
10:45:18 AgentService: fallback chain built — 12 ready   (chain includes openrouter)
18:22:56 config.toml saved with openrouter removed
18:22:57 ConfigWatcher: reloaded — firing 6 callback(s)
18:29:12 persist_sticky_pair: fallback to openrouter/…   <- still there
18:30:52 persist_sticky_pair: fallback to openrouter/…
18:31:27 persist_sticky_pair: fallback to openrouter/…
```

## Root cause

`AgentService.fallback_providers` was a plain `Vec<Arc<dyn Provider>>` populated exactly once, in `AgentService::new` via `build_fallback_providers(config)`. Nothing rebuilt it afterwards — `grep` found zero callers assigning to it outside construction and a test-only setter.

The ConfigWatcher callback in `cli/ui.rs` hot-swapped the **primary** provider (`swap_provider`) and refreshed tools, commands, approval policy and context budget. The fallback chain, the other half of the same config block, was not among the six callbacks. So the two halves of `[providers.fallback]` reloaded at different speeds: primary immediately, chain never.

This is what makes the user-visible report *"config is not hot reloading"* half true: five of six things reload; the chain is frozen at process start.

## Fix

- `fallback_providers` moves behind an `RwLock` so it can be replaced on a live service.
- `AgentService::fallback_chain_snapshot()` for readers (returns a cloned `Vec` of `Arc`s — callers walk the chain across `.await` points, so handing out a guard would both pin a stale list and make the future non-`Send`).
- `AgentService::reload_fallback_providers(&config)` rebuilds from a freshly loaded config and logs the before/after chain.
- Called from the ConfigWatcher in `cli/ui.rs`, next to the existing `swap_provider`, and unconditionally — a failed primary rebuild must not leave a stale chain behind.
- An empty rebuilt chain is applied as-is: clearing `fallback_chain` has to clear the runtime chain, otherwise removal is unrepresentable.

## Regression test

`src/tests/fallback_chain_hot_reload_test.rs` — reload replaces the chain, removal of a provider drops it from the runtime chain, and clearing the config empties it.

[...]

Issue #262, opened 2026-07-02
Config hot-reload doesn't refresh in-memory Provider state — providers retain startup-time values


leshchenko1979:
## Summary

The config file watcher in `src/utils/config_watcher.rs` correctly watches the **active profile's** home directory (e.g. `~/.opencrabs/profiles/ops/` — `opencrabs_home()` → `resolve_profile_home()` resolves to the profile dir, not `~/.opencrabs/`). The doc comment on the watcher ("Watches the `~/.opencrabs/` directory") is misleading but the code is correct.

The deeper bug is one layer down: when the watcher fires and calls `Config::load()` → `Config::set_current()`, only the `Config` struct is refreshed. The in-memory `Provider` instances (held in `ChannelFactory::provider` and elsewhere) are constructed **once at daemon startup** and never re-built. Consequence: any change to a provider's `default_model`, `enabled`, `models`, `base_url`, etc. in the profile's `config.toml` (or `keys.toml`) is silently ignored until the daemon is restarted.

This is the underlying cause of the cascading A2A Gatus-alert failures seen in the last 24h (today's two alerts both died with the same `minimax` stream-drop because flipping `opencode.enabled = true` in the file never took effect in the running daemon, so the global default provider remained the dead `minimax`).

## Repro

1. Start the daemon on the `ops` profile.
2. Confirm the global default provider is `minimax` (or whatever's first enabled in `REGISTRATIONS`).
3. Edit `~/.opencrabs/profiles/ops/config.toml` and set `[providers.opencode] enabled = true`.
4. Touch `keys.toml` or `commands.toml` to force the watcher to fire (the watcher ignores edits to files in the watched dir unless one of the three target names — so a no-op touch on the file works).
5. Check the daemon log — `ConfigWatcher: reloaded — firing N callback(s)` appears, but `provider.default_model()` and the global default resolution still reflect the **pre-edit** values.
6. `curl` the live API for the provider you just enabled: key works, models list is current. But the daemon's in-memory provider is still in its startup state.

Same effect for any of: `default_model` typo, `enabled` flag, `base_url`, custom provider rename, fallback chain re-order. All require a daemon restart to take effect.

## Root cause

`src/channels/factory.rs:95` (and the analogous sites in the A2A gateway and TUI):

```rust
AgentService::new(self.provider.clone(), self.service_context.clone(), &config)
```

[...]

PR #1720, opened 2026-09-23
fix(config): RSI and channel agents track live config; guard both freeze shapes (#1696, #1700)

### LINKED ISSUE #1696: fix(rsi): RSI engine builds its provider and fallback chain from a startup config snapshot — config changes never reach it
## Summary

`spawn_rsi_engine` clones the startup `Config` **once** and passes that clone to every RSI cycle. `run_rsi_agent_cycle` then builds the RSI provider and its `[providers.fallback]` chain from that frozen snapshot.

Result: **config changes never reach a running RSI engine.** Editing the fallback chain, deleting a provider section, rotating a key, or changing a model has no effect on RSI until the process is restarted — while the same loop, a few lines away, reads the live config mirror correctly for `rsi_enabled`.

This contradicts the rule the config module documents for itself in `src/config/current.rs`:

> Use [`Config::current`] everywhere a request/turn needs config. Reserve the disk-reading `Config::load` for exactly three places: startup seeding, the watcher itself, and write-then-reread.

## Evidence — code

`src/brain/rsi.rs` (identical on `v0.5.3` and `main`):

```rust
// rsi.rs:1205
pub fn spawn_rsi_engine(
    pool: crate::db::Pool,
    config: &Config,
    notification_tx: mpsc::UnboundedSender<RsiNotification>,
    headless: bool,
) {
    let pool_clone = pool.clone();
    let config_clone = config.clone();          // rsi.rs:1213  <-- snapshot, taken once
    tokio::spawn(async move {
        ...
        // Re-read from the live config mirror every cycle, so flipping
        // `rsi_enabled` in config.toml takes effect on the next boundary
        // without a restart, in both directions.
        if !rsi_effectively_enabled(&Config::current(), headless) {   // rsi.rs:1225  <-- LIVE
```

```rust
// rsi.rs:1809  — inside the cycle loop
match run_rsi_agent_cycle(repo.pool().clone(), &config_clone, &opportunities)
//                                          ^^^^^^^^^^^^^ FROZEN
```

`run_rsi_agent_cycle` uses that argument for both provider resolution and chain construction:

```rust
// rsi.rs:1020
crate::brain::provider::factory::create_provider_by_name(config, provider_name).await
// rsi.rs:1040
crate::brain::provider::factory::wrap_with_fallback_chain(config, provider).await?
```

Both read the passed `config` (`create_provider_by_name` resolves custom entries from `config.providers.custom`; `wrap_with_fallback_chain` reads `config.providers.fallback` and calls `create_fallback(config, name)`). Neither consults `Config::current()`.

Call site — the config handed in is the one loaded once at startup:

```rust
// src/cli/ui.rs:513 (v0.5.3) / src/cli/ui.rs:516 (main)
crate::brain::rsi::spawn_rsi_engine(db.pool().clone(), config, rsi_tx, headless);
```

`cmd_chat_inner(config: &Config, ...)` receives it from `cmd_chat`, which receives it from the single `load_config()` call on the startup path (`Config::load()`), and the clone lives for the lifetime of the engine task.

**The contradiction in one screen:** the loop reads the live mirror for the enable gate (rsi.rs:1225) and a startup snapshot for the provider + chain (rsi.rs:1809). The hot-reload machinery is already there and already used; it just is not used for the thing that costs money.

## Evidence — runtime

One live install (TUI running since 2026-09-13). At `2026-09-23T05:37:10+07:00`, the RSI engine built its chain:

```
rsi_engine: Fallback provider 'inferhub-alt' ready
rsi_engine: Fallback provider 'commandgoat'  ready
rsi_engine: Fallback provider '9router'      ready
rsi_engine: Fallback provider 'openrouter'   ready
rsi_engine: Wrapping primary provider 'inferhub' with 4 fallback(s)
```

At that moment the on-disk config was:

```toml
[providers.fallback]
providers = ["modelstudio", "zai-coding"]   # 2 entries, not 4
```

and the `commandgoat` section **did not exist at all** — it had been deleted hours earlier. The only remaining occurrences of that string in `config.toml` are two unrelated comments. The engine nonetheless built a live provider from it and logged `Fallback provider 'commandgoat' ready`.

Same snapshot, second symptom:

```
2026-09-22T23:04:59.039634+07:00 WARN rsi_engine: src/brain/rsi.rs:1024:
RSI: self_improvement_provider 'zai_coding' failed to create
(unknown provider 'zai_coding': not a built-in ...)
```

The frozen snapshot still carried a spelling that the live config had already moved past.

Cost, from the same day's log: **203** RSI fallback attempts into providers that no longer exist in config, each walking the retry ladder (`Retry attempt 1/4 after 952ms for ...`) before giving up. The RSI session `d91ef0f8` accumulated 5,879 log lines.

## Why this is a bug, not intended behaviour

- The config module documents `Config::current()` as the correct accessor for anything a turn needs, and reserves `Config::load` for startup seeding.
- `spawn_rsi_engine`'s own doc comment says the engine re-reads the live mirror so config changes apply "without a restart" — true for `rsi_enabled`, false for everything else in the same function.
- A user editing the fallback chain reasonably expects it to apply. The engine instead retries deleted providers indefinitely, and the failure is silent from the user's side: RSI cycles keep running and keep reporting, they just waste their retry budget first.

## Impact

- Config edits require a restart to reach RSI — undocumented, and invisible because nothing errors.
- Dead providers stay in the RSI chain forever, burning retry time every cycle.
- Deleted provider sections keep being constructed (and their stale keys, if any, keep being used).
- Key rotation does not reach RSI until restart.
- Any future `self_improvement_model` / provider change silently misses the RSI path.

## Proposed fix

Resolve the config per cycle instead of capturing it at spawn:

```rust
// rsi.rs:1809
let config_now = Config::current();
match run_rsi_agent_cycle(repo.pool().clone(), &config_now, &opportunities).await
```

and drop `config_clone` (keep the `&Config` parameter for the startup path if it is still needed elsewhere in the task, or read `Config::current()` once at the top of the loop body).

A regression test in the style of the existing `src/tests/rsi_fallback_wrap_test.rs` would pin this: mutate the fallback chain between two cycle invocations and assert the second cycle observes the new chain.

## Reproduction

1. Start the TUI with a `[providers.fallback]` chain of N entries.
2. Change `providers.fallback.providers` on disk (e.g. remove an entry, or delete a provider's section entirely) — the config watcher reloads it for normal sessions.
3. Wait for an RSI cycle and inspect the log: the engine still logs `Fallback provider '<removed>' ready` and wraps the old chain length.

## Environment

- OpenCrabs `v0.5.3` (bug also present on `main` at `43f1c465`, same line numbers)
- Linux x86_64, TUI mode
- Verified against upstream source, not a local tree

## Lineage (added 2026-09-24, when the fix landed)

This is not a one-off snapshot. It is the half of the #1249 sweep that #1249 never
reached, and the body used to cite nothing, which is exactly how a closed family
keeps looking finished:

| Link | What it is |
|---|---|
| #1249 | Ancestor. Its fix is `34b26778f`; `git show --name-only 34b26778f \| grep -E "rsi\|channels"` is empty. It wired `reload_fallback_providers` into exactly one production caller, the TUI service in `src/cli/ui.rs`. |
| PR #166 | Origin of this snapshot (`b35341747`, "fix: RSI agent now respects [providers.fallback] chain"). It gave the RSI cycle its fallback chain *through* the startup `Config` clone, so the chain has been frozen at process start since the day RSI gained fallback support. |
| #1700 | Sibling. Same freeze, different shape: channel agents reach their provider through `ChannelFactory`'s bare `Arc<dyn Provider>` field rather than through a `Config`. |
| #262 | Grandparent: "Config hot-reload doesn't refresh in-memory Provider state". |
| #1720 | Tracking PR. Fixes #1696 and #1700 as one class and closes both. |

The class is not "someone cloned a `Config`". It is: a long-lived task or struct
holds something DERIVED from one config read past the next change notification.
Derived means a `Config` clone (here), an `Arc<dyn Provider>`, or a fallback chain
(#1700). A census on `Config` alone passes while the channel half stays frozen,
which is why #1720 adds `src/tests/config_freeze_class_test.rs` guarding both shapes.

One deviation from the requested regression test, stated rather than dropped
quietly: this issue asks for a cycle-mutation test in the style of
`rsi_fallback_wrap_test.rs`. It is not buildable hermetically as written, because
`build_fallback_providers` requires `config.providers.fallback` to be `Some` and
enabled and silently skips any name whose `create_provider_by_name` fails, so a
test without real credentials cannot construct a two-entry chain to mutate between
cycles. #1720 pins the RSI half with a source-shape guard (no `config_clone`, no
`Config` parameter on the spawner, a `Config::current()` read in the window feeding
the cycle call) and pins the chain-mutation behaviour for real on the channel half,
where a `ChannelFactory` and its live agents build against an in-memory DB.

### LINKED ISSUE #1700: fix(channels): fallback chain never hot-reloads for channel agents, so a provider deleted from fallback_chain keeps serving traffic until restart
## Summary

`AgentService::reload_fallback_providers` (#1249) is wired only into the TUI's agent service. Channel agents (Telegram/WhatsApp/Discord/Slack) are **separate `AgentService` instances** created by `ChannelFactory::create_agent_service_full` and never receive the reload, so their fallback chain stays frozen at the config from process start, or from whenever the channel was spawned.

Effect: a provider **deleted from `[providers.fallback].providers`** keeps receiving live traffic from every channel session until the process restarts.

This is the channel half of #262 ("Config hot-reload doesn't refresh in-memory Provider state"). #262 fixed the **primary** provider (channels now call `swap_provider` / `swap_provider_for_session`). #1249 fixed the **chain**, but only for `app.agent_service()`. The two halves are asymmetric: channel files call `swap_provider` 4-6 times and `reload_fallback_providers` **zero** times.

## Environment

- OpenCrabs v0.5.3 (running binary), reproduced against `main` @ `43f1c465`
- Telegram channel, 4 forum topics, 1 TUI session, long-lived process (spawned 2026-09-12)
- `[providers.fallback].providers` was edited several times while the process ran

## Repro (observed, not constructed)

1. Process running, channel agents spawned. At **2026-09-21T06:39:16** the log shows `ChannelManager: spawning Telegram bot (2 allowed users)`. The chain in config at that time was `[inferhub-alt, commandgoat, 9router, openrouter]`.
2. At **2026-09-22T13:14:57** `commandgoat` and `9router` are removed from `fallback_chain`. The TUI's service logs the reload:

   ```
   ConfigWatcher: fallback chain reloaded, [inferhub-alt, openrouter]
                                     (was [inferhub-alt, commandgoat, 9router, openrouter])
   ```

3. Channel sessions keep dispatching to the **removed** providers:

   ```
   2026-09-22T15:43:43  session_id=80dedc98…  Fallback 'commandgoat': running its configured model 'z-ai/glm-5.3-flash'
   2026-09-22T18:07:50  session_id=cc1582fe…  Fallback 'commandgoat': running its configured model 'z-ai/glm-5.3-flash'
   2026-09-22T22:41:45  session_id=c58ce982…  Fallback 'commandgoat': running its configured model 'z-ai/glm-5.3-flash'
   2026-09-23T07:33:07  session_id=ff442a3e…  Fallback 'commandgoat': running its configured model 'z-ai/glm-5.3-flash'
   2026-09-23T08:45:44  session_id=80dedc98…  Fallback '9router':     running its configured model 'smart-free'
   ```

   18 dispatches to the removed `commandgoat`/`9router` across 4 Telegram sessions, spanning **19h+ after removal**, the last one while this issue was being written.

4. Config live at that moment: `["modelstudio", "zai-coding"]`, neither of which appears in any of those dispatches.

5. Contrast: the TUI session (`44c44939`) never dispatched to a removed provider after the reload, because its service is the one the ConfigWatcher actually reaches.

## Why

`reload_fallback_providers` has exactly **3 call sites in the whole repo**: one production, two tests.

```bash
$ gh search code --repo adolfousier/opencrabs "reload_fallback_providers" --json path
src/brain/agent/service/builder.rs          # the definition
src/cli/ui.rs                               # the ONLY production call site
src/tests/fallback_chain_hot_reload_test.rs # tests
```

The single production call site targets one service:

```rust
// src/cli/ui.rs:1784 (v0.5.3) / :1819 (main)
let agent = app.agent_service().clone();
...
// src/cli/ui.rs:1813 (v0.5.3) / :1848 (main)
agent.reload_fallback_providers(&cfg).await;
```

Channel agents are different instances, built per channel:

```rust
// src/channels/factory.rs:159-166 (v0.5.3, identical in main)
pub async fn create_agent_service_full(...) -> Arc<AgentService> {
    let config = self.config_rx.borrow().clone();
    let mut builder =
        AgentService::new(self.provider.clone(), self.service_context.clone(), &config)
```

```rust
// src/channels/manager.rs:195-197 (v0.5.3)
let tg_agent_service = self
    .channel_factory
    .create_agent_service_full(Some(reaction_cb), Some(enqueue_cb))
    .await;
```

The chain is built once in the constructor and lives behind a lock that only that instance can replace:

```rust
// src/brain/agent/service/builder.rs:521-522 (v0.5.3) / :526-527 (main)
fallback_providers: std::sync::RwLock::new(
    Self::build_fallback_providers(config).await,
),
```

And the doc comment on the field states the invariant that the channel path violates:

```rust
// src/brain/agent/service/builder.rs:415-425 (v0.5.3)
/// Behind a lock since #1249: the chain used to be built once at
/// construction and never touched again, so editing `fallback_chain` in
/// config.toml had no effect until a full restart. The ConfigWatcher
/// hot-swapped the PRIMARY provider and left this list frozen, which is how
/// a provider deleted from the chain kept serving turns for hours.
```

Per-channel files swap the **primary** but never the chain. Grep on `main`:

| file | `swap_provider` | `reload_fallback_providers` |
|---|---|---|
| `src/channels/telegram/agent.rs` | 4 | **0** |
| `src/channels/discord/agent.rs` | 4 | **0** |
| `src/channels/slack/handler.rs` | 4 | **0** |
| `src/channels/commands.rs` | 6 | **0** |
| `src/channels/factory.rs` | 0 | **0** |

The ConfigWatcher *does* broadcast the new config to channels over the `config_tx` watch channel (`ui.rs:1788-1792`), and channels already hold `config_rx`. The config **reaches** them; nothing acts on it for the chain.

## Impact

- **Silent misrouting.** Every stale dispatch is a request to a provider the user deleted. Some are dead (401 "api key revoked", 401 "Invalid 'Authorization' header"), some are live-but-unwanted (403 free-tier, 429 quota), and the chain walks them before landing somewhere usable.
- **Latency.** Each stale entry costs a full retry ladder before the next entry is tried. Observed chains of `inferhub-alt` → `commandgoat` → `9router` → `openrouter`, all failing, before recovery.
- **Worse on long-lived daemons.** The channel agent is spawned once and lives for weeks. Its chain is the config from spawn time, so the drift window equals the process uptime.
- **No error surface.** Nothing logs "this chain is stale". The only way to notice is to correlate `Fallback 'X': running its configured model` against the current config.

## Proposed fix

Give the channel agents the same reload the TUI service gets. Two shapes, either works:

**A. Per-channel handler (smallest diff).** Channels already receive the new config on `config_rx`. In the channel message loop, on a config change, call `reload_fallback_providers` on the agent service next to the existing per-session `swap_provider`:

```rust
// alongside the existing agent.swap_provider_for_session(...) in the config-change arm
agent.reload_fallback_providers(&cfg).await;
```

**B. Central (better for many channels).** Have `ChannelManager` keep an `Arc<AgentService>` per running channel, and let the ConfigWatcher iterate them the way it already iterates `app.agent_service()`.

Regression test: follow `src/tests/fallback_chain_hot_reload_test.rs`, but assert against a service built via `ChannelFactory::create_agent_service_full`, i.e. the test must fail if only the TUI path reloads.

## Checklist

- [x] I searched for existing issues before opening this one (checked #262, #1249, #1696, and 10 keyword searches; none covers the channel half)
- [x] I am running the latest release (v0.5.3)
- [x] I can reproduce this consistently (4 Telegram sessions, 18 dispatches to removed providers, last one 2026-09-23T08:45:44)

## Notes

- The same class of bug was reported separately for the RSI engine, which clones the config once at app start: #1696. Different snapshot site (RSI clones the whole `Config`; channels hold a per-instance `RwLock` that nobody writes), same symptom.
- #262 and #1249 are both closed; neither covers the channel half.

adolfousier:
## What this fixes

Two open issues, one class of bug. The class is: **a long-lived task or struct
holds something derived from one config read past the next change notification.**
The derived value is what freezes: a `Config` clone in #1696, an
`Arc<dyn Provider>` and a fallback chain in #1700.

#1249 closed the sweep for one path only. Its fix (`34b26778f`) touches no RSI
file and no channel file, and `reload_fallback_providers` had exactly one
production caller (`src/cli/ui.rs`, on `app.agent_service()`). That is why the
sweep could look finished four times while both of these stayed frozen.

Closes #1696
Closes #1700

## Root cause, per fragment

| # | Fragment | Frozen thing | How it got there | Why #1249 missed it |
|---|----------|--------------|------------------|---------------------|
| 1696 | `spawn_rsi_engine` | the whole `Config`, then provider + fallback chain built from it | PR #166 (`b35341747`) wired `wrap_with_fallback_chain` through the startup clone | the reload was never wired into the RSI cycle at all |
| 1700 | `ChannelFactory` | a bare `Arc<dyn Provider>` field built once, plus every `AgentService` it hands out | the factory has always cached its provider | the only reload caller was the TUI service |
| 1700 (unnamed in the issue) | A2A agent, WhatsApp/Trello channels, secondary-profile cron factory | the provider `Arc` they received from the factory | they hold it through the factory, not through a `Config` | a census on `Config` alone passes while all of these stay frozen |

### Correction to #1700's premise

The issue states that #262 fixed the channel **primary**. It did not, not in the
reload sense: there is **zero** `.changed()` consumer of `config_rx` anywhere in
the repo, and all 28 channel config reads are `borrow()` per message. The
`swap_provider` calls in `src/channels/telegram/agent.rs:826-827` and
`:1091-1092` belong to the in-chat `/models` handler, not to a reload path.
WhatsApp and Trello have 0 swaps and 0 reloads. So the primary half was frozen
too, and this PR fixes both halves in one place.

## What changed

| Commit | Change |
|--------|--------|
| `b3a89ca1b` | `src/brain/rsi.rs`: startup snapshot deleted. `spawn_rsi_engine` no longer takes a `Config` at all, so the freeze cannot be reintroduced by accident. One live `Config::current()` read per cycle boundary (rsi.rs:1823), deliberately not seven scattered reads, so a save landing mid-cycle cannot tear the provider against its chain. |
| `bdbf3aa17` | `src/channels/factory.rs`: provider slot is now `RwLock<Arc<dyn Provider>>`; the factory keeps `Weak` refs to every `AgentService` it builds and gains `reload_providers(config, primary)`. `src/cli/ui.rs`: the watcher rebuilds the primary **once** and hands the same `Arc` to the TUI service and the factory registry, so they cannot diverge. The chain reload runs even when the rebuild fails (`None`), per the #1249 rule. |
| `b12dce122` | `src/tests/fallback_chain_hot_reload_test.rs`: 4 behavioural tests through a real `ChannelFactory` (chain cleared, primary swapped, agents built after a reload inherit the reloaded primary, dropped agents pruned and the rest still reloaded). `builder.rs`: `set_fallback_providers_for_test` widened `&mut self` to `&self`; the field is an `RwLock` and never needed exclusive access. |
| `679aac2a2` | `src/tests/config_freeze_class_test.rs`: class guard covering both shapes, plus a self-test of the detector itself so the guard cannot pass by matching nothing. |

No lock is held across an `.await` in `reload_providers`: the registry guard ends
with its `match` block, and the `Vec<Arc<AgentService>>` is taken out before the
iteration.

## Tests

Re-proved on tip `679aac2a2` (tree clean), 2026-09-24 00:53-00:58 UTC, with the
CI-matching commands from `.github/workflows/ci.yml`:

```
cargo fmt --all -- --check                                  exit 0, empty log
cargo clippy --locked --lib --bins --tests --examples \
  --all-features -- -D warnings                             exit 0, 0 warnings
cargo test --all-features config_freeze_class               6 passed; 0 failed
cargo test --all-features fallback_chain_hot_reload         7 passed; 0 failed
cargo test --all-features session_provider_wrap             9 passed; 0 failed
```

Every receipt is read from the log's `test result:` line, not from an exit file
alone. The full cross-platform suite stays CI's job; it is the gate below.

Mutation receipts (the tests must fail when the fix is undone):

| Mutation | Result |
|----------|--------|
| `reload_providers` loop body deleted | 3 tests fail |
| `provider.write()` slot assignment neutered | exactly 1 test fails |

## Deviation from #1696's requested test, stated openly

The issue asks for "a regression test in the style of `rsi_fallback_wrap_test.rs`:
mutate the fallback chain between two cycle invocations and assert the second
cycle observes the new chain." That is not buildable hermetically:
`build_fallback_providers` requires `config.providers.fallback` to be `Some` and
enabled, and it silently skips any name whose `create_provider_by_name` fails, so
a test with no real provider credentials cannot construct a two-entry chain to
mutate between cycles. The RSI half is therefore pinned by a source-shape guard
(`rsi_engine_takes_no_config_and_reads_it_live`: no `config_clone`, no `Config`
parameter on the spawner, and a `Config::current()` read in the window feeding the
cycle call), while the chain-mutation behaviour the issue wants is pinned for real
on the channel half, where a `ChannelFactory` and its live agents can be built
against an in-memory DB. The asymmetry is in the test file's module doc comment.

## Merge gate

Owner authorized the merge on green CI (2026-09-24 00:04 UTC). Gate is the CI
workflow's `Test` job on this exact sha, not "a green run exists". Merged with
`gh pr merge --rebase --admin` (linear history, no squash on this repo).

Unfiled fragments this PR does NOT touch, recorded so they are not lost: A2A
agent built at `src/cli/ui.rs:2003` from the factory (now reloads via the
registry, but has no independent test), WhatsApp/Trello zero-swap channels, the
secondary-profile cron factory whose watch sender is dropped at
`src/cli/ui.rs:195-199`, and `/models` persistence: `switch_model`
(`src/channels/commands.rs:2162`) writes nothing to disk while its error path
returns `"Model saved but failed to reload provider"` (`:2195`), which saves
nothing.


adolfousier:
## Fix ledger: commits, files, receipts

Narrative and root-cause table are in the PR body. This is the copy-pasteable
ledger for whoever verifies it next.

### Commits (linear, one concern each, `origin/main` = `c1e691c77`)

| SHA | Concern | Files |
|-----|---------|-------|
| `b3a89ca1b` | #1696: RSI startup snapshot deleted | `src/brain/rsi.rs`, `src/cli/ui.rs` (call site) |
| `bdbf3aa17` | #1700: channel factory reloads its live agents | `src/channels/factory.rs`, `src/cli/ui.rs` (watcher) |
| `b12dce122` | #1700: behavioural tests for the channel half | `src/tests/fallback_chain_hot_reload_test.rs`, `src/tests/session_provider_wrap_test.rs`, `src/brain/agent/service/builder.rs` |
| `679aac2a2` | Class guard for both shapes | `src/tests/config_freeze_class_test.rs` (new), `src/tests/mod.rs`, `src/channels/factory.rs` (doc pointer) |

Branch head: `679aac2a246416912d700671d6fb26c1e8345019` (confirmed by
`git ls-remote origin refs/heads/fix/config-freeze-1696-1700`).

### Gate receipts, all on macOS arm64 against the branch head

| Gate | Command | Receipt |
|------|---------|---------|
| Lint (all targets, CI shape) | `arch -arm64 cargo clippy --all-features --all-targets -- -D warnings` | exit 0, 0 errors, 0 warnings |
| Class guard | `arch -arm64 cargo test --all-features config_freeze_class` | `test result: ok. 6 passed; 0 failed; 9196 filtered out` |
| Channel reload | `arch -arm64 cargo test --all-features fallback_chain_hot_reload` | `test result: ok. 7 passed; 0 failed; 9195 filtered out` |

Last write to any branch file was 00:40:35; the lint receipt is 00:41:35 and the
guard receipt 00:42:53, so both postdate the tree they certify.

### Mutation evidence (the tests must fail when the fix is undone)

| Mutation | Expected | Actual |
|----------|----------|--------|
| `reload_providers` loop body deleted | the 3 live-agent reload tests fail | `FAILED. 4 passed; 3 failed` |
| `provider.write()` slot assignment neutered | the post-reload-build test fails alone | `FAILED. 6 passed; 1 failed` |

### Two implementation notes that are easy to trip over

1. `set_fallback_providers_for_test` had to widen from `&mut self` to `&self`.
   `Arc::get_mut` requires `weak_count == 0`, so registering the `Weak` refs made
   the old seam permanently unusable on an `Arc<AgentService>`, which is the only
   shape production holds. The backing field is an `RwLock` and never needed
   exclusive access; 9 redundant `mut` bindings across two test files went with it.
2. The watcher rebuilds the primary **once** and passes the same `Arc` to the TUI
   service and to `factory.reload_providers`, so the two cannot diverge. The chain
   reload runs even when the rebuild fails (`primary == None`), per the #1249 rule
   that a broken rebuild must not strand the old chain.

No lock is held across an `.await` in `reload_providers`: the registry guard ends
with its `match` block and the `Vec<Arc<AgentService>>` is extracted before
iteration.

### Merge

Held. Needs explicit approval plus green CI.

